PR #10, opened 2016-08-30
build: add missing arvconfig header

dbcesar:


EmmanuelP:
Thanks

EmmanuelP:
It broke introspection, so I have moved the declaration in the ARAVIS_HDRS_NO_INTRO block. Please let me know if it still fixes your issue.

EmmanuelP:
I have reverted the installation of arvconfig.h in 0b1a532caf9ab55b7567d305303dcc5b5cea54a5, as it breaks distcheck.

But I have also removed a reference to this file in arvtypes.h, so the installation of aravis should be ok.

Please open a new issue in case of problem.

PR #12, opened 2016-09-01
Adds a variable to add custom defines to the configure script

Ippo343:
The user can set the ARAVIS_DEFINES environment variable to pass
additional definitions. The definitions are concatenated to the
ARAVIS_CFLAGS variable.

The rationale is to allow the use of custom defines without changing
the default ARAVIS_CFLAGS computed by the script.


EmmanuelP:
I don't think it is needed, you should be able to define additional flags by setting the right environment variable at configure time. See ./configure --help for an overview of what is available.

Ippo343:
I looked into the configure script and I agree this is not really needed, I'll use the configure script.
Sorry for bothering you :)

PR #13, opened 2016-09-01
Fast heartbeat

Ippo343:
This pull request adds a compilation flag called "FAST_HEARTBEAT" that causes the redefinition of some heartbeat related values in order to produce a much faster heartbeat than the default.

This depends on pull request #12


src/arvgvdevice.h
```diff
@@ -29,11 +29,33 @@
 
 G_BEGIN_DECLS
 
+#ifdef FAST_HEARTBEAT
```
EmmanuelP: Defines should be prefixed ARV_, or ARAVIS if they come from the configure script.

In this case, ARAVIS_FAST_HEARTBEAT should be defined by ./configure. See how it is done for ARAVIS_BUILD_USB or ARAVIS_BUILD_VIEWER in configure.ac .

src/arvgvdevice.h
```diff
@@ -29,11 +29,33 @@
...
+    #define        ARV_GV_DEVICE_GVCP_N_RETRIES_DEFAULT    3
+    #define        ARV_GV_DEVICE_GVCP_TIMEOUT_MS_DEFAULT   25
+    #define        ARV_GV_DEVICE_HEARTBEAT_PERIOD_US       50000
+    #define        ARV_GV_DEVICE_HEARTBEAT_RETRY_DELAY_US  1000
+    #define        ARV_GV_DEVICE_HEARTBEAT_RETRY_TIMEOUT_S 0.25
+#endif
+
+#ifndef	ARV_GV_DEVICE_GVCP_N_RETRIES_DEFAULT
```
EmmanuelP: A single #if #else #endif sequence should be used here, instead of this multiple #ifndef #define #endif sequences.

EmmanuelP:
Thanks.

Ippo343:
No, thank you!

PR #18, opened 2016-09-07
print more descriptive error if user interface file can't be loaded

flixr:
Instead of printing that the user interface file is missing, print the actual error.

E.g. on Ubuntu 14.04 with gtk+ 3.10 it will now print the more informative message:
```
Cant't load user interface file ('/usr/local/share/aravis-0.6/arv-viewer.ui'): /usr/local/share/aravis-0.6/arv-viewer.ui: required gtk+ version 3.12, current version is 3.10
```

### DIFF
--- a/viewer/arvviewer.c
+++ b/viewer/arvviewer.c
@@ -127,7 +127,7 @@ typedef struct {
 	gboolean flip_horizontal;
 
 	double exposure_min, exposure_max;
-	
+
 	NotifyNotification *notification;
 
 	GtkWidget *main_window;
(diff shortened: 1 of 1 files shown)


viewer/arvviewer.c
```diff
@@ -1117,8 +1117,9 @@ activate (GApplication *application)
 
 	ui_filename = g_build_filename (ARAVIS_DATA_DIR, "arv-viewer.ui", NULL);
 
-	if (!gtk_builder_add_from_file (builder, ui_filename, NULL))
-		g_error ("The user interface file is missing ('%s')", ui_filename);
+	GError *err = NULL;
```
EmmanuelP: Variable declaration must happen at the beginning of the function.

err must be freed using g_error_free.

EmmanuelP:
Thanks.

PR #19, opened 2016-09-08
Allows to enable the fast heartbeat flag from the configuration script

Ippo343:
This adds a new option to the configure script (--enable-fast-heartbeat) that defines the ARAVIS_FAST_HEARTBEAT in src/arvgvdevice.h thus enabling the much faster heartbeat option.

refs #13


EmmanuelP:
I have pushed your commit.

I took the time to think about this feature. A proper final implementation would be to adjust these heartbeat parameters by taking into account the value of the heartbeat register ARV_GVBS_HEARTBEAT_TIMEOUT_OFFSET (0x938).

Ippo343:
Thank you very much! I'll try to find the time to take a proper look into that register.
Merci beaucoup!

PR #20, opened 2016-09-29
build: check for libcap-ng for packet-socket

hfiguiere:
it didn't have libcap-ng along libaudit, but still tried to build with packet-socket.


EmmanuelP:
Thanks.

PR #32, opened 2016-11-18
Updated the required version of glib

blaz-kranjc:
glib-2.0 > 2.34 is required as g_clear_pointer() is used.
https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-clear-pointer

### DIFF
--- a/src/arvmisc.h
+++ b/src/arvmisc.h
@@ -106,6 +106,31 @@ ArvPixelFormat 	arv_pixel_format_from_gst_0_10_caps 		(const char *name, int bpp
 
 #endif
 
+#if GLIB_CHECK_VERSION(2,34,0)
+
+#define arv_g_clear_pointer(pp, destroy) arv_g_clear_pointer(pp, destroy)
+
+#else
+
+#define arv_g_clear_pointer(pp, destroy)                                   \
+  G_STMT_START {                                                               \
+    G_STATIC_ASSERT (sizeof *(pp) == sizeof (gpointer));                       \
+    /* Only one access, please */                                              \
+    gpointer *_pp = (gpointer *) (pp);                                         \
+    gpointer _p;                                                               \
+    /* This assignment is needed to avoid a gcc warning */                     \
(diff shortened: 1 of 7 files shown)


blaz-kranjc:
AC_CHECK_DECL() on TPACKET_V3 should probably also be added to quickly fail with compilation on older kernels (< 3)

EmmanuelP:
Hi,

I should not have used this function in the first place. Anyway, the right fix would be to add a compatibility replacement function in arvmisc, like what is done for g_mutex*, and leave the glib requirement as it is.

blaz-kranjc:
Another function used that is not compatible with the older version of glib is g_hash_table_contains.
(https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-contains)

I'll check what is done for g_mutex.

EmmanuelP:
> Another function used that is not compatible with the older version of glib is g_hash_table_contains.
(https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-contains)

This one can probably be replaced by a call to g_hash_table_lookup_extended.

src/arvmisc.h
```diff
@@ -106,6 +106,31 @@ ArvPixelFormat 	arv_pixel_format_from_gst_0_10_caps 		(const char *name, int bpp
 
 #endif
 
+#if GLIB_CHECK_VERSION(2,34,0)
+
+#define arv_g_clear_pointer(pp, destroy) arv_g_clear_pointer(pp, destroy)
```
EmmanuelP: #define arv_g_clear_pointer(pp, destroy) g_clear_pointer(pp, destroy)

blaz-kranjc:
Should be OK now.

EmmanuelP:
Pushed to master. I have changed the g_clear_pointer patch to make it use the same name. It will avoid compilation issues if  g_clear_pointer is accidentally used in the futur.

PR #36, opened 2016-12-23
Feature/usb3 ximea

mcguire-steve:
Emmanuel, this PR adds support for emitting trigger signals via GPIO pins on a camera. The use case in my lab is a stereo pair, where one camera (the master) emits a signal that triggers its slave camera. At least on my Ximea XiQ cameras, it appears that a freerunning camera does not emit FrameStart. However, when the camera is set to software triggering, it does emit the signal. I'm not sure if this behavior is standards-compliant or not. This work was done without reference to any AIA documentation.
Cheers,
Steve

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ On some platforms (like Ubuntu), you may have to configure the dynamic linker (l
 Using the GNU build system on Mac OS X is not directly supported, but can be mimicked by augmenting the install procedure above with some environment settings:
 
 ```
-brew install gettext intltool
+brew install gettext intltool gtk-doc
 brew link --force gettext
 aclocal
 autoconf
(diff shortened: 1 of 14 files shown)


EmmanuelP:
Could you please rebase your work on master ?

Thanks.

mcguire-steve:
@EmmanuelP , I think the rebase is complete as requested.

EmmanuelP:
@mcguire-steve the history of this pull request is so complex it is very hard to review, an it includes too much unrelated changes in different part of aravis.

I have commited to trigger API extension to master, with some documentation and coding style fixes.

Please rework this pull request in at least two new ones: one for the GPIO part, and another for unpause. The history should be clean and linear starting from current master.

Regarding the GPIO part, take care to respect the current coding style: spaces, variable declaration at the beginning of the code blocks, variable naming (nSources -> n_sources), function naming consistency.

As for unpause, there should not be device specific code in generic functions in arvcamera. If something specific is needed for a device type, it should be in the corresponding source files (arvuv*.[ch] in the case of unpause).

Thanks for your work.

Cheers.

EmmanuelP:
On last thing: there is a commit that adds a comment about realtime-tests not working on OSX. It would be better to fix it instead of indicating a workaround in README. Could you or @crheckman file a bug explaining what goes wrong on OSX ? Thanks.

PR #37, opened 2016-12-27
fix memory leak in GC feature node

mdavidsaver:
valgrind flags a "definite" leak in arv_gc_feature_node_set_attribute() from g_strdup().  Free this string in feature node's finalize.


EmmanuelP:
Merged in master.

Thanks a lot.

PR #40, opened 2017-01-02
fake_camera: detect zero frame period

mdavidsaver:
Otherwise results in SIGFPE if client configures with zero frame rate.

Since zero period isn't likely to be an intended condition (at least it wasn't for me) emit a warning and fallback to 1Hz.


EmmanuelP:
Merged into master.

Thanks.

PR #41, opened 2017-02-03
Add claim of both interface[0] and interface[1]

mcguire-steve:
Resolves runtime kernel message of unclaimed interface use from libusb, since the control pipes used for GenICam data and register setting are on interface 0, while the image data is retrieved from an endpoint on interface 1


EmmanuelP:
I have pushed a extended version of your patch. Thanks.

PR #42, opened 2017-02-03
Feature/uv unpause

mcguire-steve:
Add feature to start UV image acquisition in a paused state. On MacOS, issuing a libusb_bulk_transfer on a handle without data wedges libusb because of a timeout. Future reads fail even after data is available (i.e. after starting acquisition) - requiring the device to be unplugged and re-enumerated.

@EmmanuelP , regarding the add of device-specific code into arvcamera, my goal was to come up with a way to solve the problem of inadvertently reading from a pipe before acquisition was started (thereby wedging MacOS) while preserving device independence in higher-level code. I'm certainly open to a redesign to meet those goals.

### DIFF
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -650,6 +650,15 @@ arv_camera_start_acquisition (ArvCamera *camera)
 {
 	g_return_if_fail (ARV_IS_CAMERA (camera));
 	arv_device_execute_command (camera->priv->device, "AcquisitionStart");
+
+	if (ARV_IS_UV_DEVICE(camera->priv->device))
+	  {
+	    ArvUvDevice *uv_cam;
+	    uv_cam = ARV_UV_DEVICE(camera->priv->device);
+	    ArvUvStream *uv_stream;
+	    uv_stream = arv_uv_device_get_stream(uv_cam);
+	    arv_uv_stream_unpause(uv_stream);
+	  }
 }
 
 /**
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -38,6 +38,7 @@
  * </example>
  */
 
+#include <arvconfig.h>
 #include <arvcamera.h>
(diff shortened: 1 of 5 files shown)


EmmanuelP:
If it is really necessary to unpause, I would prefer if it was implemented in ArvDevice. ArvCamera is an helper library that provides a friendly API for the most common features, but it should always be possible to use a camera by only using the ArvDevice API.  By putting device specific code in generic functions of ArvCamera like start_acquisition, it breaks ArvDevice.

I think the best solution would be to implement 2 new functions in ArvDevice: arv_device_start_acquisition and arv_device_stop_acquisition, and virtuals in ArvDevice that would (before_start and after_stop) that would be called by theses functions. ArvUvDevice would implement before_start and call unpause. arv_camera_start_acquisition and stop_acquisition would be simple wrapper around arv_device_start/stop.

mcguire-steve:
For reference, this error manifests on MacOS Sierra 10.12.3 with libusb 1.0.21. Thanks for the design suggestions - I'll refactor to comply.

EmmanuelP:
Hi Steve,

Did you have a chance to work on this patch ?

mcguire-steve:
Emmanuel, I admit that other projects have had priority recently - can I have more time? I'm trying to defend my PhD soon :)

EmmanuelP:
Don't worry Steve, I was just asking. Phd first.

EmmanuelP:
Closed by Mistake.

EmmanuelP:
Aravis has changed a lot since this pull request. There is now a start/stop thread API that could be used to implement this functionality if needed.

Also, it is a bit puzzling it is not possible to recover from a timeout, which is the main reason of this pull request.

So let's close this pull request as the changes don't make sense with respect to the current aravis code.

@mcguire-steve please open an issue if you are still experiencing timeouts.

PR #44, opened 2017-02-07
Fake camera controller privilege fixups

mdavidsaver:
Three changes related to handling of Control Channel Privilege register in the fake camera.

The main change is to correctly handle a client which explicitly relinquishes control privilege by writing CCP=0.  This needs to clear ```gv_camera->controller_address``` so the the write permission check works correctly.

I found this when writing some test code which needs to pass control between two clients: some setup code, and the code under test.  With these changes, this hand off works correctly.

I will note that when ```!write_access``` it would be better to return a permission denied error (I think involving ```ARV_GVCP_PACKET_TYPE_ERROR```).  However, I couldn't find an easy way to do this.

### DIFF
--- a/src/arvfakegvcamera.c
+++ b/src/arvfakegvcamera.c
@@ -359,7 +359,7 @@ handle_control_packet (ArvFakeGvCamera *gv_camera, GSocket *socket,
 			(time.tv_nsec - gv_camera->controller_time.tv_nsec) / 1000000;
 
 		if (elapsed_ms > arv_fake_camera_get_heartbeat_timeout (gv_camera->camera)) {
-			g_object_ref (gv_camera->controller_address);
+			g_object_unref (gv_camera->controller_address);
 			gv_camera->controller_address = NULL;
 			write_access = TRUE;
 			arv_warning_device ("[FakeGvCamera::handle_control_packet] Heartbeat timeout");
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Thanks Michael.

EmmanuelP:
> I will note that when !write_access it would be better to return a permission denied error (I think involving ARV_GVCP_PACKET_TYPE_ERROR). However, I couldn't find an easy way to do this.

You are right. I don't remember what real cameras return in this case. If you have a real camera and are familiar with wireshark, you could try to get a capture of a failed write attempt.

mdavidsaver:
> you could try to get a capture of a failed write attempt.

When a second client attempts to write CCP=2

> 0000   42 01 00 82 00 08 ff 30 00 00 0a 00 00 00 00 02

A TIS DMK 33GX174 responds with

> 0000   80 06 00 83 00 00 ff 30

So I guess in ```arv_gvcp_packet_new_write_register_ack()``` it would be necessary to (later) replace ```ARV_GVCP_PACKET_TYPE_ACK``` with ```ARV_GVCP_PACKET_TYPE_ERROR``` and set size==0.

FYI, wireshark decodes 0x8006 as ```GEV_STATUS_ACCESS_DENIED```

PR #49, opened 2017-03-02
Single threaded image acquisition mode for GigEVision and USB3Vision

nerian-vision:
I have added an option to perform single-threaded image acquisition for GigEVision and USB3Vision. On low-performance embedded computers the single threaded processing causes significant performance improvements. I have noticed drastic framerate increases on an embedded ARM board (5 fps to 20 fps for GigE and 15 to 30 fps for USB)

The single-threaded mode is optional and activated through the following device options:

`ARV_GV_STREAM_OPTION_THREADING_DISABLED` for GigEVision
`ARV_UV_STREAM_OPTION_THREADING_DISABLED` for USB3Vision

Please note that I had to add the method `arv_camera_uv_set_stream_options()` for setting the USB3Vision options.

If this option is set, the processing has to be scheduled manually. This is done by calling one of the following methods, depending on whether the given camera is a GigE or USB3 camera:

`arv_gv_stream_schedule_thread()`
`arv_uv_stream_schedule_thread()`

To realizing this functionality for USB, I had to switch the code to the asynchroneous libusb API.


EmmanuelP:
Thanks for your contribution. I will review your patch soon. Some quick comments:

- I'd like to have a simple program in order to test your patch (which would allow to test both single and multi-threaded acquisition).
- given you have implemented single threaded acquisition for both GigEVision and USB3Vision, it would be better if *_schedule_thread was implemented in arv_stream.

nerian-vision:
Thanks for the feedback. I will put some further work into this.

nerian-vision:
The schedule function has been moved to `arv_stream` and a test program called `singlethread-test` has been added to tests/. I also fixed a bug in the USB implementation.

The new test program can run in multi- or single-threading mode, depending on whether you pass a 0 or 1 as command line argument.

While testing I noticed that `arv_stream_pop_buffer()` is not yet working. This function will block forever, as the schedule function is not called. Currently, only `arv_stream_try_pop_buffer()` works for single threading. I'm not sure if its worth the effort finding a solution for making `arv_stream_pop_buffer()` work, as using this function contradicts the motivation for implementing the single thread mode, which was to have a non-blocking processing loop in a single thread process.

I tested GigEVision with an old Basler Scout and USB3Vision with a Basler dart and a PointGrey/FLIR Chameleon3.

EmmanuelP:
Please be patient for the review. I don't have much time currently :(

Thanks.

EmmanuelP:
No complete review yet, but I have tried the sample program.

It is hard to compare the performance of the 2 modes, as the busy loop in main makes the process use 100% of a CPU. That is fixed by adding a usleep call. But still, even with this sleep, the USB3Vision thread still consume 100% of a CPU with your patches.

nerian-vision:
There was indeed a bug in the USB3Vision multi-thread mode, where the libusb calls were non-blocking instead of blocking. I just fixed that. I guess on a PC there won't be much of a performance difference. But on a slow ARM board this makes quite a difference.

nerian-vision:
I added a new features which allows blocking calls to the schedule function (by adding a new boolean parameter). This removes the need for a call to sleep in the test program.

EmmanuelP:
I have to say I'm not completely convinced by this serie of patches.

I adds complexity to a part of aravis that is already a bit too complex to my taste, especially in the GigE Vision case. And for the platforms used by most of the users (x86_64), it should have almost no impact on performances.

Even in the case of an ARM platform, I still don't understand why the use of a thread for handling data would cause such a performance hit. I'd like to be sure the difference you see doesn't come from elsewhere. I will try to do some tests using the raspberry board I have. But I don't know when I will be able to find time for that.

Meanwhile, don't hesitate to post figures you may collect testing your work.

nerian-vision:
The real issue is the Linux scheduler which is often configured to have 20 ms time slices or similar. If there is only one CPU core available and the acquisition thread is running on the same core as the onward processing, then this can have quite bad impacts on the timing.

This also means that the thread waiting for the image data has to frequently sleep to allow the image acquisition thread to process. A call to sleep has very nondeterministic timing, as it depends on the scheduler when processing is resumed. For lowest latency, a busy loop would be the preferred solution.

Currently, using a busy loop will drastically slow down the image acquisition thread on the same core. This is the scenario that I was using when I reported the observed frame rate increases (5 fps to 20 fps for GigE and 15 to 30 fps for USB). This might not be a common situation for a desktop application but typical for a real-time embedded application.

Allowing the user to control the image acquisition thread also has another advantage: The user is able to adjust the thread properties such as thread priority or cpu affinity.

EmmanuelP:
> Currently, using a busy loop will drastically slow down the image acquisition thread on the same core. This is the scenario that I was using when I reported the observed frame rate increases (5 fps to 20 fps for GigE and 15 to 30 fps for USB). This might not be a common situation for a desktop application but typical for a real-time embedded application.
> 
> Allowing the user to control the image acquisition thread also has another advantage: The user is able to adjust the thread properties such as thread priority or cpu affinity.

Did you try to use the stream callback parameters ?

This callback should be called from the receiving thread at thread initialization, thread termination, buffer start and buffer completion (should be, because it seems the related code is missing in arvuvstream, I'm going to fix that).

When using this callback, you should be able to change the receiving thread priority and cpu affinity (In our realtime application, I'm making this thread hard realtime using RTAI), and it will behave almost like the single thread mode.

nerian-vision:
I haven't tried the callbacks and I'm actually just using USB3Vision at the moment. Anyway, I'm not going to add multi-threaded code to my control loop. I have made way too many bad experiences with that. I might maintain this patch set separately.

PR #51, opened 2017-03-10
emit_signals synchronization

### LINKED ISSUE #50: Race in stream new-buffer signal leads to self destruction
It seems that ```g_signal_emit()``` manipulates the ref. count of ```ArvStream``` instances.  This means that it is possible for the stream thread to have the last reference, and to attempt to destroy the stream.  This triggers an assert() failure from ```g_thread_join()``` when the stream thread attempts to join itself.

This is fairly difficult to trigger.  I've encountered it twice in ~2 weeks of destruct testing involving triggering cleanup while still streaming frames.

The symptom is

```
(process:23937): GLib-ERROR **: file /build/glib2.0-y6934K/glib2.0-2.42.1/./glib/gthread-posix.c: line 1218 (g_system_thread_wait): error 'Resource deadlock avoided' during 'pthread_join (pt->system_thread, NULL)'
```

A stack trace of the faulting thread is.

```
#0  0x00007fe5374f2d30 in g_logv () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007fe5374f2f6f in g_log () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007fe53753039e in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007fe537512c95 in g_thread_join ()
   from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#4  0x00007fe537a39fee in arv_gv_stream_finalize (object=0x7fe4ec0c62d0)
    at arvgvstream.c:1188
#5  0x00007fe5377c626a in g_object_unref ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#6  0x00007fe5377db494 in g_signal_emit_valist ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#7  0x00007fe5377db9df in g_signal_emit ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#8  0x00007fe537a387df in _close_frame (
    thread_data=thread_data@entry=0x7fe4ec081960, 
    frame=frame@entry=0x7fe51001dc00) at arvgvstream.c:499
#9  0x00007fe537a38fb2 in _check_frame_completion (
    current_frame=<optimized out>, time_us=1488981195843477, 
    thread_data=0x7fe4ec081960) at arvgvstream.c:540
#10 _loop (thread_data=0x7fe4ec081960) at arvgvstream.c:715
#11 arv_gv_stream_thread (data=0x7fe4ec081960) at arvgvstream.c:933
#12 0x00007fe537512845 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#13 0x00007fe53728d0a4 in start_thread (arg=0x7fe4a7bf8700)
```

mdavidsaver:
Attempt to fix #50.  Add a mutex to ```_ArvStreamPrivate``` which is locked in ```arv_stream_push_output_buffer()``` around the test of ```emit_signals``` and the call to ```g_signal_emit()```.  Also in ```arv_stream_set_emit_signals()``` when ```emit_signals``` is updated.

The stream shutdown sequence should include setting emit_signals=0, which by locking the mutex would ensure that ```g_signal_emit()``` isn't running after ```arv_stream_set_emit_signals()``` returns.

With some hesitation, print a warning if a stream is finalized w/ emit_signals=true.  Hopefully, this will give some hint about avoiding this a deadlock without being too annoying.

I've only tested the glib < 2.32 condition.

### DIFF
--- a/src/arvstream.c
+++ b/src/arvstream.c
@@ -53,9 +53,9 @@ struct _ArvStreamPrivate {
 	GAsyncQueue *input_queue;
 	GAsyncQueue *output_queue;
 #if GLIB_CHECK_VERSION(2,32,0)
-	GMutex mutex;
+	GRecMutex mutex;
 #else
-	GMutex *mutex;
+	GRecMutex *mutex;
 #endif
 
 	gboolean emit_signals;
@@ -189,18 +189,18 @@ arv_stream_push_output_buffer (ArvStream *stream, ArvBuffer *buffer)
 	g_async_queue_push (stream->priv->output_queue, buffer);
 
 #if GLIB_CHECK_VERSION(2,32,0)
-	g_mutex_lock (&stream->priv->mutex);
+	g_rec_mutex_lock (&stream->priv->mutex);
 #else
-	g_mutex_lock (stream->priv->mutex);
+	g_rec_mutex_lock (stream->priv->mutex);
 #endif
 
 	if (stream->priv->emit_signals)
 		g_signal_emit (stream, arv_stream_signals[ARV_STREAM_SIGNAL_NEW_BUFFER], 0);
 
 #if GLIB_CHECK_VERSION(2,32,0)
-	g_mutex_unlock (&stream->priv->mutex);
+	g_rec_mutex_unlock (&stream->priv->mutex);
 #else
-	g_mutex_unlock (stream->priv->mutex);
+	g_rec_mutex_unlock (stream->priv->mutex);
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


EmmanuelP:
>  With some hesitation, print a warning if a stream is finalized w/ emit_signals=true. Hopefully, this will give some hint about avoiding this a deadlock without being too annoying.

Well, I don't like that either, but I did not find a better idea. We could try to keep a ref on stream as long as the thread is running, using g_object_add_toggle_ref. That means we could know when the thread is the owner of the last reference, and then destroy the stream object without joining itself.

But that would still mean a thread would be running after the user as released the last reference he knows.

I guess enforcing emit_signal = false on finalization is the best we could do.

EmmanuelP:
Pushed to master.

Thanks.

PR #59, opened 2017-04-04
Save system time for each captured frame

nerian-vision:
I have made an extension that saves the system time at which a frame was recorded, in addition to the already recorded camera time. The timestamp is saved to the buffer as soon as the leader is received, in order to minimize delays. There are two new functions for getting and setting the system timestamp:

`arv_buffer_get_system_time()`
`arv_buffer_set_system_time()`

The system time is saved in microseconds, which is unlike the camera timestamp that is saved in nanoseconds. This is because the used `g_get_real_time()` function only has microsecond resolution. If this makes the interface inconsistent, it could also be converted to nanoseconds.

### DIFF
--- a/src/arvbuffer.c
+++ b/src/arvbuffer.c
@@ -330,7 +330,7 @@ arv_buffer_set_timestamp (ArvBuffer *buffer, guint64 timestamp_ns)
  */
 
 guint64
-arv_buffer_get_system_time (ArvBuffer *buffer)
+arv_buffer_get_system_timestamp (ArvBuffer *buffer)
 {
 	g_return_val_if_fail (ARV_IS_BUFFER (buffer), 0);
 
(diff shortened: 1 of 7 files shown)


src/arvbuffer.c
```diff
@@ -318,6 +318,46 @@ arv_buffer_set_timestamp (ArvBuffer *buffer, guint64 timestamp_ns)
...
+ * Gets the system timestamp for when the frame was received. Expressed in
+ * microseconds.
+ *
+ * Returns: buffer timestamp, in microseconds.
+ *
+ * Since: 0.4.0
+ */
+
+guint64
+arv_buffer_get_system_time (ArvBuffer *buffer)
```
EmmanuelP: Please rename these functions arv_buffer_[get|set]_system_timestamp.

src/arvbufferprivate.h
```diff
@@ -46,6 +46,7 @@ struct _ArvBufferPrivate {
 
 	guint32 frame_id;
 	guint64 timestamp_ns;
+    guint64 systemtime_us;
```
EmmanuelP: systemtime_us -> system_timestamp_ns

nerian-vision: I made the requested changes.

EmmanuelP:
Commited with some documentation related changes.

Thanks.

PR #64, opened 2017-04-28
Adds automake to list of brew install programs

markrickert:
I found that after running `brew install gettext intltool gtk-doc` I was missing the `aclocal` command. Installing `automake` with brew solves this issue.


EmmanuelP:
Thanks Mark.

PR #66, opened 2017-05-09
Create Usb device by name "Manufacturer;Product;Serial"

thecaptury:
this is similar to the gige method that can instantiate by IP address.

### DIFF
--- a/src/arvgvinterface.c
+++ b/src/arvgvinterface.c
@@ -512,9 +512,9 @@ _open_device (ArvInterface *interface, const char *device_id)
 
 		memset(&hints, 0, sizeof (hints));
 		hints.ai_family = AF_INET;
-		hints.ai_socktype = SOCK_STREAM;
+		hints.ai_socktype = SOCK_DGRAM;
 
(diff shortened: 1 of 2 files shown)


thecaptury:
The thing is that I want to avoid using arv_update_device_list() because it is slow for GigE and I want to use DHCP if possible anyway. Using the device serial number is acceptable but I think it makes sense to check again whether the device is there now. So an alternative implementation would look something like this:

```
if (device_infos == NULL) {
    GList *device_list;

    _discover(uv_interface, uv_interface->priv->device_ids);
    device_infos = g_hash_table_lookup (uv_interface->priv->devices, device_id);
    if (device_infos == NULL)
        return NULL;
}
```

EmmanuelP:
arv_uv_interface_open_device  already calls _discover. What would your alternative implementation add to the existing code ?

thecaptury:
Ok. You are right, using "Manufacturer-Serial" works. reject this pull request please.

thanks.

EmmanuelP:
Ok.

Regarding GigEvision, I have just pushed a patch that makes opening a device faster. The one second delay is gone.

PR #69, opened 2017-05-12
UV: already set payload size and count before stream creation

nerian-vision:
Currently the payload size and payload count are only set after the first stream is created. This seems to have caused USB errors on one system. I have thus moved the setting of payload size and count from arvuvstream to avuvdevice. I have also kept the setting of these parameters in arvuvdevice in case they might change after the image acquisition parameters have been set.


EmmanuelP:
You have also changed the computations. Could you explain what you have done ?

EmmanuelP:
> This seems to have caused USB errors on one system.

Which systems ?

EmmanuelP:
I have tested your patch. After a few seconds, using a The Imaging Source camera, arv-viewer starts to complain about USB timeouts. A reboot of the machine is needed to regain access to the camera.

nerian-vision:
My primary test system currently is a Zynq Ultrascale+ SoC, which includes an ARM Cortex A53 CPU and integrated USB3 controller.

Seems like the issue I was having was primarily a Kernel problem (fixed in https://github.com/torvalds/linux/commit/e210c422b6fdd2dc123bedc588f399aefd8bf9de). After patching the kernel, it seems like the the changes are not necessary, so I guess the pull request can be closed.

I changed the computations, as I think the original computation was wrong. However, both versions produce the same result. To my understanding, the payload size is the size of a single transfer. Hence the the overall size of the burst without the last transfer should be si_payload_size*si_payload_count. As you configure the camera to transmit only one transfer per burst, this change shouldn't actually make a difference. However, if the payload count should change in future versions, then this change should be necessary.

I'm still having quite some stability problems on the ARM platform, particularly when receiving image data from two cameras at a the same time. I tend to get a lot of corrupted frames, but they seem to be corrupted on the host, as they contain chunks of pixel data from the respective other camera. I realized that reducing the payload size to the requested payload size, rather than the maximum, reduces this problem, but it's still there. I'll keep on debugging...

PR #72, opened 2017-05-19
UV: bugfix for initial leader / trailer size of 0 for some Point Grey…

nerian-vision:
Some Point Grey cameras seem to report a required leader and trailer size of 0 after plugging in. So the first attempt to access them after plugging in always fails. On the second attempt they report the correct size. I observed this with CM3-U3-13S2M-CS and BFLY-U3-05S2M-CS.

I increased the bulk transfers to the maximum transfer size for the leader and trailer, which seems to fix the problem.

### DIFF
--- a/src/arvuvstream.c
+++ b/src/arvuvstream.c
@@ -90,12 +90,12 @@ arv_uv_stream_thread (void *data)
 		transferred = 0;
 
 		if (buffer == NULL)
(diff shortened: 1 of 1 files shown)


EmmanuelP:
I have pushed a modified version of your patch, that adds some debugging messages. Thanks.

PR #77, opened 2017-06-27
gvcp: flush socket receive buffers before sending new commands

ids-imaging:
This fixes a problem that occurs when the camera is under high pressure
and not feasible to acknowledge gvcp requests immediately
(Or the gvcp_timeout_ms is to low).

Description:

A gvcp command was sent by _read_memory(), _write_memory(),
_read_register() or _write_register(). Waiting for the acknowledge
packet (with the corrosponding packet id) times out.
After that the camera sends this acknowledge package which will not be
processed by now. Therefore it remains in the socket receive buffer.

Impact:

Following gvcp commands will also fail because they get the wrong
acknowledge package
(the old one with wrong packed id => unexpected answer).
Its not possible to recover from that state. Old acknowledge packages
will allways remain in the receive buffer.
Therefore gvcp commands will fail by now.

The _read_register() call in arv_gv_device_heartbeat_thread() will also
fail in each iteration.
For this reason io_data->is_controller will be set to FALSE and the
camera connection will be lost after heartbeat timeout.

Solution:

Flush socket receive buffers before sending new gvcp commands.
Old received acknowledges will be discarded.

### DIFF
--- a/src/arvgvdevice.c
+++ b/src/arvgvdevice.c
@@ -122,6 +122,22 @@ static GRegex *arv_gv_device_url_regex = NULL;
 	return arv_gv_device_url_regex;
 }
 
+static void
+_flush_socket_buffer(ArvGvDeviceIOData *io_data)
+{
+	GError *local_error = NULL;
+	gboolean success = TRUE;
+
+	while (success &&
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Thanks.

PR #81, opened 2017-06-30
build: use builddir for arvenumtypes.{c,h}

jvff:
The build was failing when attempting to use a separate build directory. For example:

    mkdir build
    cd build
    ../configure
    make

The cause appears to be that all files passed to introspection were prefixed with the source directory, so it was assuming that all files were present in the source directory. This was not the case for the generated files: arvenumtypes.c and arvenumtypes.h, which are generated in the *build* directory.

This commit separates the generated files from the source files and prefixes them to the build directory and the source directory, respectively.


EmmanuelP:
Pushed to master and aravis-0-4 branch.

Thanks.

PR #119, opened 2017-11-16
Update Mac build instructions to fix errors

RaphiePS:
I got various errors from `./configure` and `make`, and needed to add these commands to get it all running smoothly.


EmmanuelP:
Pushed to master.

Thanks.

Issue #126, opened 2018-01-08
UV stream thread loses sync with the device when there is no destination buffer available


TIS-Tim:
After applying my proposed patch for issue #125, the situation can arise that there are no buffers available for arv_uv_stream_thread to copy data to.

If arv_stream_pop_input_buffer returns NULL, the thread stays in the state where it expects a leader, while the device sends image data.

Because the payload data is larger than the requested size for the leader, libusb then reports LIBUSB_TRANSFER_OVERFLOW, which, according to its documentation, should be avoided.

I think only asking for a leader to be transferred if a buffer is available to be filled would be a solution to avoid this situation.

EmmanuelP:
> I think only asking for a leader to be transferred if a buffer is available to be filled would be a solution to avoid this situation.

That was the previous situation. I have fixed the issue by requesting a maximum transfer size when we are waiting for a leader. Thanks a lot for the report.

Cheers.

PR #128, opened 2018-01-09
Return buffers to aravis only after gstreamer no longer references them

TIS-Tim:
Testing showed that the gstreamer pipeline can hold more than one buffer, but arv-viewer currently assumes that the pipeline contains exactly one. It returns the buffer to aravis before gstreamer is done with it, leading to corruption of the displayed image with new data.

This change uses the notification callback from GstBuffer to return the buffer to aravis, which should be the right point in time.

### DIFF
--- a/viewer/arvviewer.c
+++ b/viewer/arvviewer.c
@@ -241,8 +241,30 @@ arv_viewer_value_from_log (double value, double min, double max)
 	return pow (10.0, (value * (log10 (max) - log10 (min)) + log10 (min)));
 }
 
+struct free_arv_buffer_cb_param
+{
+	GWeakRef stream;
+	GWeakRef arv_buffer;
+};
+
+static void
+free_arv_buffer_cb (void* param)
+{
+	struct free_arv_buffer_cb_param* cb_param = (struct free_arv_buffer_cb_param*)param;
+
+	ArvStream* stream = g_weak_ref_get (&cb_param->stream);
+	ArvBuffer* arv_buffer = g_weak_ref_get (&cb_param->arv_buffer);
+
+	if( stream && arv_buffer )
+		arv_stream_push_buffer( stream, arv_buffer );
+
+	g_weak_ref_clear( &cb_param->stream );
+	g_weak_ref_clear( &cb_param->arv_buffer );
+	g_free( param );
+}
+
 static GstBuffer *
-arv_to_gst_buffer (ArvBuffer *arv_buffer)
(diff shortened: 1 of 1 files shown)


viewer/arvviewer.c
```diff
@@ -241,8 +241,30 @@ arv_viewer_value_from_log (double value, double min, double max)
...
+	GWeakRef stream;
+	GWeakRef arv_buffer;
+};
+
+static void
+free_arv_buffer_cb (void* param)
+{
+	struct free_arv_buffer_cb_param* cb_param = (struct free_arv_buffer_cb_param*)param;
+
+	ArvStream* stream = g_weak_ref_get (&cb_param->stream);
```
EmmanuelP: I did not know this API, thanks !
 Reading the documentation, g_weak_ref converts the weak reference to a normal reference. That probably means there is a reference leak in this function. You should call g_object_unref line 260.

viewer/arvviewer.c
```diff
@@ -241,8 +241,30 @@ arv_viewer_value_from_log (double value, double min, double max)
...
+	ArvBuffer* arv_buffer = g_weak_ref_get (&cb_param->arv_buffer);
```
EmmanuelP: arv_buffer should always be valid in this callback, so using g_weak_ref on it is not needed. You should be careful to unref it in the case arv_stream_push_buffer is not called.

viewer/arvviewer.c
```diff
@@ -289,11 +313,19 @@ new_buffer_cb (ArvStream *stream, ArvViewer *viewer)
...
 
-	if (viewer->last_buffer != NULL)
-		arv_stream_push_buffer (stream, viewer->last_buffer);
-	viewer->last_buffer = arv_buffer;
+		gst_app_src_push_buffer (GST_APP_SRC (viewer->appsrc), arv_to_gst_buffer (arv_buffer, cb_param));
+
+		viewer->last_buffer = arv_buffer;
+	}
+	else
+	{
```
EmmanuelP: } else {

TIS-Tim:
I updated my branch with your proposed changes. Good catch on the missing unref, that was actually a bad regression.

EmmanuelP:
It looks mostly fine for a merge to master, but I need to perform some tests before.

Thanks Tim.

EmmanuelP:
Sorry for the time it takes, but I still did not take the time to review and test your patches.

EmmanuelP:
Pushed to master with a reference leak fix and debug output. Thanks.

I was not sure about the patches because after a while, the TIS camera became unresponsive. Then I realized its was due to the issue #126 you have already reported.

PR #131, opened 2018-01-27
Complete the installation instructions

rehno-lindeque:
Fixes #101

PR #132, opened 2018-01-28
Delete orphaned sentence in README

rehno-lindeque:
Following up on #131 (thank you for the clarification there). I believe this piece of text made it into the final commit by accident.


EmmanuelP:
Right. Thanks.

PR #136, opened 2018-02-16
add second argument 'precision' to ROUND function

rickvavilin:
Second argument 'precision' to ROUND function added as described by GenICam Standard (page 39)


src/arvevaluator.c
```diff
@@ -508,6 +508,13 @@ arv_get_next_token (char **expression, gboolean previous_token_was_operand, gboo
 	return token;
 }
 
+double round_with_precision(double x, gint64 precision)
```
EmmanuelP: Make it static in order to not leak function outside of arvevaluator.c
Please use the coding style

src/arvevaluator.c
```diff
@@ -508,6 +508,13 @@ arv_get_next_token (char **expression, gboolean previous_token_was_operand, gboo
...
+{
+	double pow_multiplier = pow(10, precision);
+	return (round (x*pow_multiplier))/pow_multiplier;
+}
+
+
```
EmmanuelP: No double empty lines, please

src/arvevaluator.c
```diff
@@ -807,7 +813,9 @@ evaluate (GSList *token_stack, GHashTable *variables, gint64 *v_int64, double *v
 							      ceil (arv_value_get_double (&stack[index])));
 				break;
 			case ARV_EVALUATOR_TOKEN_FUNCTION_ROUND:
-				arv_value_set_double (&stack[index], round (arv_value_get_double (&stack[index])));
+				arv_value_set_double (&stack[index-1], round_with_precision (arv_value_get_double (&stack[index-1]),
+																			 arv_value_get_int64 (&stack[index])));
+
```
EmmanuelP: No double empty lines, please

src/arvevaluator.c
```diff
@@ -542,7 +549,6 @@ evaluate (GSList *token_stack, GHashTable *variables, gint64 *v_int64, double *v
 		}
 
 		arv_evaluator_token_debug (token, variables);
-
```
EmmanuelP: Leave the empty line

EmmanuelP:
Sorry, I did not check the documentation before my comment to your patch. In fact, the second parameter is optional, with a default value set to 0. The expression evaluator of aravis doesn't support optional arguments. Which means this pull request needs more work in order to not break the support of ROUND with one argument, which is the most common case.

rickvavilin:
Ok, i'll try to implement optional arguments support later. I need this, because i use aravis with Teledyne Dalsa Genie Nano which use ROUND with two parameters.

EmmanuelP:
That would be great. Thanks.

rickvavilin:
Hello, where i can read the coding style?

EmmanuelP:
Well, the coding style is not documented, but any source file in src should be conform.

rickvavilin:
Hello, I implemented optional arguments support.

EmmanuelP:
Thanks. I'll have a look asap.

Did you run the test suite ?

```
make check
cd tests
make check-valgrind-memcheck
```

rickvavilin:
Test suite show problems, i'll fix it.
Can you describe how to run test suite in README?

rickvavilin:
The tests is all green now.

EmmanuelP:
I'm no fond of the handling of the parenthesis level while evaluating. It should be done while parsing.
I have found a discussion on the subject of implementing the parsing of functions with a variable number of arguments here:

https://blog.kallisti.net.nz/2008/02/extension-to-the-shunting-yard-algorithm-to-allow-variable-numbers-of-arguments-to-functions/

In the end of the discussion, there seem to be interesting implementation suggestions.

rickvavilin:
Should both handling parentheses level and calculating count of arguments be done while parsing? Or it possible to count of arguments while evaluating?

EmmanuelP:
I think it is more efficient to do it while parsing, as it would not be repeated at each evaluation.
There is different approaches discussed on the linked page.

rickvavilin:
I read the discussion on linked page. All discussed approaches suppose some work at the evaluation stage. So I remove parenthesis handling from evaluation, but leave arguments counting.

EmmanuelP:
That is not what I expected, I would have preferred if the argument counting was in the parsing function.
But the feature is implemented and all the tests pass, so I'm pushing it to master.
Thanks for your work on this.

PR #139, opened 2018-03-13
Manpages

cmarmo:
Hi,
I'm packaging aravis for Debian: at least basic man pages for executable tools are required.
I've created them via `help2man` then modified a little because of the  missing `--version` option.
If you think that it could be useful for the official distribution too here is the code of my patch.
`arv-test` has no `--help` option too, so there is no man page for it.
Best,
Chiara


EmmanuelP:
Pushed to master.

Thanks.

PR #140, opened 2018-03-13
Updating viewer appstream metadata

cmarmo:
HI,
Debian complains about outdated appstream metadata.
Here is the version validated via `appstreamcli` (AppStream version: 0.11.8).
If you think it can be useful for the original package too...
Best,
Chiara


EmmanuelP:
Merged to master. Thanks Chiara.

PR #153, opened 2018-05-03
Expose the raw ticks timestamp of packets.

malban:
This commit exposes the raw ticks timestamp in ArvGvspPackets to ArvBuffers.

Background:

I'm trying to get the actual camera timestamp from a Prosilica GT camera that supports PTP time synchronization, but the ArvBuffer always seems to report system time.  

Looking into it, I found that the timestamp_tick_frequency is not successfully read in arv_gv_device_get_timestamp_tick_frequency, resulting in system time being used as a fallback.

This commit stores and exposes the raw ticks timestamp in the buffer object so that it is available even when the clock frequency is unknown or unavailable.

For this camera the frequency can be obtained by querying "GevTimestampTickFrequency" through arv_device_get_integer_feature_value, which I can use to convert the ticks timestamp to an actual time.

### DIFF
--- a/docs/reference/aravis/aravis-sections.txt
+++ b/docs/reference/aravis/aravis-sections.txt
@@ -123,6 +123,7 @@ arv_buffer_get_user_data
 arv_buffer_get_data
 arv_buffer_get_chunk_data
 arv_buffer_get_timestamp
+arv_buffer_get_timestamp_ticks
 arv_buffer_set_timestamp
 arv_buffer_get_system_timestamp
 arv_buffer_set_system_timestamp
@@ -1752,6 +1753,7 @@ arv_gvsp_packet_get_packet_type
 arv_gvsp_packet_get_content_type
 arv_gvsp_packet_get_payload_type
 arv_gvsp_packet_get_pixel_format
+arv_gvsp_packet_get_timestamp_ticks
 arv_gvsp_packet_get_timestamp
 arv_gvsp_packet_get_width
 arv_gvsp_packet_get_x_offset
--- a/src/arvbuffer.c
+++ b/src/arvbuffer.c
@@ -299,6 +299,26 @@ arv_buffer_get_timestamp (ArvBuffer *buffer)
 	return buffer->priv->timestamp_ns;
 }
 
+
+/**
+ * arv_buffer_get_timestamp_ticks:
+ * @buffer: a #ArvBuffer
+ *
+ * Gets the buffer camera timestamp, expressed as ticks.
+ *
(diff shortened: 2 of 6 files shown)


EmmanuelP:
Hi,

Thanks for the pull requests.

The problems I have with this PR are:
  - It is incomplete, and should also implement raw ticks for UBS and fake devices
  - On USB devices, timestamp is delivered as nanoseconds, so raw ticks makes no sense
  - It is a workaround for an issue that should be fixed

Currently, timestamp tick frequency is read directly using the register address, without using a genicam feature:
https://github.com/AravisProject/aravis/blob/a648012bc8bf377a3e5aad3b4798dd6463b14d7e/src/arvgvdevice.c#L711
I guess it explains why with your device this value is not read correctly (while I don't understand why the timestamp tick frequency register address would be different on your device).

So the correct fix would be to change arv_gv_device_get_timestamp_tick_frequency and make it use the GevTimestampTickFrequency. As this feature is not always defined in the device genicam data, a default definition should be added here:
https://github.com/AravisProject/aravis/blob/a648012bc8bf377a3e5aad3b4798dd6463b14d7e/src/arvgvdevice.c#L1012

Could you try to implement this change ?

malban:
Yeah, I'll take a look.

malban:
This is the XML for the GevTimestampTickFrequency:

```
<Integer Name="GevTimestampTickFrequency" NameSpace="Standard">
  <Extension>
   <PvRename Name="TimeStampFrequency"/>
  </Extension>
  <ToolTip>Frequency of image timestamp, in Hertz.</ToolTip>
  <Description>Frequency of image timestamp, in Hertz.</Description>
  <pValue>RegTimestampTickFreqLow</pValue>
</Integer>

 <IntReg Name="RegTimestampTickFreqLow" NameSpace="Custom">
  <Address>0x940</Address>
  <Length>4</Length>
  <AccessMode>RO</AccessMode>
  <pPort>Device</pPort>
  <Sign>Unsigned</Sign>
  <Endianess>BigEndian</Endianess>
</IntReg>
```
So, it looks like it is accessing the same register as ARV_GVBS_TIMESTAMP_TICK_FREQUENCY_LOW_OFFSET in arv_gv_device_get_timestamp_tick_frequency. 

Presumably this camera only has the low part of the frequency and the failure is when the high part is read.  Is an acceptable solution to modify arv_gv_device_get_timestamp_tick_frequency to  only apply the high part if it exists?

malban:
Looking into it some more, the tick frequency is actually read in correctly and the issue with the stamp was on my end.

PR #155, opened 2018-05-05
Czech and Slovak translations

multiflexi:
I updated Czech and created Slovak translation. I am not sure about the head which was probably created automatically by some software, I changed it as I thought is right but IDK.


EmmanuelP:
Merged in master.

Thanks a lot.

PR #157, opened 2018-05-11
get rid of libcap-ng dependency

thecaptury:
Instead of asking libcap-ng whether we have permission to open raw_packet sockets try to open one. If it works great if not then we probably don't have the required permissions.


EmmanuelP:
Thanks.

PR #158, opened 2018-05-13
PointGrey is now FLIR

thecaptury:


EmmanuelP:
Pushed to master with a minor change. Thanks.

PR #193, opened 2018-09-05
gvstream: add API to free queue owned and packet receiver thread owne…

werner-Z:
API function (for GigE cams) to free all the buffers, not only the one in the input and output queues, but also the one handled by the packet receiving htread.
Reason: In our current application we have to handle frequent changes of the read-out-region for a grab in a grab-preparation phase some time before the effective image aquisition starts. Some times it can happen that an already prepared snap is aborted before capture and in this case the buffers, which never will be grabbed in, have to be removed.

### DIFF
--- a/src/arvgvstream.c
+++ b/src/arvgvstream.c
@@ -755,9 +755,18 @@ _loop (ArvGvStreamThreadData *thread_data)
 			frame = NULL;
 
 		_check_frame_completion (thread_data, time_us, frame);
+
+		// signal state of owned buffers
+		if (thread_data->frames == NULL) {
+			arv_stream_receiver_buffers_freed (thread_data->stream, TRUE);
+		}
+		else {
+			arv_stream_receiver_buffers_freed (thread_data->stream, FALSE);
+		}
 	} while (!g_atomic_int_get (&thread_data->cancel));
 
 	g_free (packet);
+	arv_stream_receiver_buffers_freed (thread_data->stream, TRUE);
 
 }
 
--- a/src/arvgvstream.c
+++ b/src/arvgvstream.c
@@ -726,6 +726,7 @@ _loop (ArvGvStreamThreadData *thread_data)
 	size_t read_count;
 	int timeout_ms;
 	int n_events;
+	int i;
 
 	arv_debug_stream ("[GvStream::loop] Standard socket method");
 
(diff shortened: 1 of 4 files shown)


EmmanuelP:
Hi Werner,

Thanks for your work on this improvement.

There are 2 issues in your patch:

  * There is a race condition where arv_stream_free_buffers may incorrectly think no buffer are used by the stream receiving thread.
  * arv_stream_free_buffers should always work.The right approach would be to signal the stream receiving thread it should release all the buffers it owns, and wait until it is actually done.

werner-Z:
Hi Emmanuel,
I fixed the potential race between arv_stream_free_buffers() and the gvstream receiver thread. I also had to improve the responsiveness of the receiver thread on a "free buffers" request (looping g_poll with reduced timeout) in case no acquisition is active. This is necessary for our application to reduce the latency when changing the read-out-region (otherwise we had 1s in worst case). Currently I implemented a basic 5ms timeout for the poll in this case. I didn't recognized any impact on the CPU load caused by the looping.

werner-Z:
Hi Emmanuell,

when do you plan to integrate this pull request into aravis:master?

EmmanuelP:
Hi Werner,

I don't have time currently to look into this pull request. Please be patient...

  Cheers,

    Emmanuel.

EmmanuelP:
Hi Werner,

I'm sorry I have wasted your time on this issue, but, as I had some time lately to devote to aravis work, I have realized I have worked on a similar feature back in 2015, without finishing it.

So I have resurrected most of this work, with some improvements, here: https://github.com/AravisProject/aravis/tree/wip/wakeupnew

There is a test program arv-roi-test that stress tests the buffer deletion.

It looks like everything works fine in the GigEVision case. I'm still working on USB3Vision support.

Please let me know if it works for your use case.

Sorry again.

EmmanuelP:
Finally, I have taken another approach, where I start/stop the receiving thread. It is easier to implement in the USB3Vision case, and the performance hit related to the destruction/creation of the thread should be small enough wrt the time the camera takes to restart the streaming.

The resulting work is in this branch: https://github.com/AravisProject/aravis/tree/wip/thread-reset

I will push it to master soon, but meanwhile I would appreciate if you could test it. Have a look at arvroitest.c to understand how the thread start/stop function should be used.

EmmanuelP:
I have merged thread-reset branch in master.

Cheers.

werner-Z:
Hi Emmanuel,
I've tested the thread-reset method and it works perfectly for our application. Thank you for your work!

bye

EmmanuelP:
Hi Werner.

Thanks for the feedback.

Cheers.

PR #198, opened 2018-10-09
Auto packet size from interface mtu

thecaptury:
Testing packet sizes larger than the mtu of the bound interface doesn't make sense. So we get the MTU with an ioctl() and use that as the starting point.
This guesses the maximum packet size correctly at the first try in most well-configured situations. Although I've seen slightly larger packet sizes working than configured.

Unfortunately, we have to iterate over all interfaces to get the name of the interface if we have only the IP address. This makes the code somewhat more ugly...


thecaptury:
Hi,

I was half-expecting you to reject it. The code is quite nasty but for me it reduces the number of tries from around 7 to 1.

No worries. I'll keep it in our branch anyway.


nils

> On 10 October 2018 at 10:23 Emmanuel Pacaud <notifications@github.com> wrote:
> 
> 
> Hi,
> 
> While I understand your point, MTU check code increase the complexity of this function (network related code always is) and it doesn't decrease the number of iteration sufficiently to be interesting.
> 
> Sorry but I'm not going to pull this patch.
> 
> Thanks anyway.
> 
> Cheers.
> 
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/AravisProject/aravis/pull/198#issuecomment-428482413

--
Dr. Nils Hasler
+49 174 175 66 42
www.TheCaptury.com
The Captury GmbH, Saarbruecken, Germany
Amtsgericht Saarbrücken, HRB 101327
Executive Director: Nils Hasler

PR #213, opened 2018-11-28
avoid infinite loop in auto packet size

thecaptury:
If arv_device_set_integer_feature_value(device, "GevSCPSPacketSize", current_size) fails for some reason current_size will stay the same and the surrounding loop will never terminate. This patch adds a check that terminates the loop if current_size does not change from one loop iteration to the next.

An alternative to this patch would be to limit the number of outer loop iterations.


EmmanuelP:
Applied. Thanks.

Issue #209, opened 2018-11-22
Wrong SI_maximum_trailer_size and SI_maximum_leader_size


mvxe:
I had a strange issue with my Allied Mako U-130B camera. Both the Aravis viewer and my own program using the Aravis API would not get anything frames via the stream unless the camera was open at least once before using the Allied Vimba API. If the camera gets unplugged and replugged or if I simply set a reset command, then it wouldn't work until it is again opened by the proprietary API at least once.

Not wanting to deal with the proprietary API as its buggy as hell, I used Wireshark to sniff out whats wrong. Apparently there are a lot of these errors:
Status: U3V_STATUS_SI_PAYLOAD_SIZE_NOT_ALIGNED (0xa003)
Status: U3V_STATUS_SI_REGISTERS_INCONSISTENT (0xa004)
And the like. It seams that the Aravis API sets the registers SI_maximum_trailer_size and SI_maximum_leader_size to values 32 and 65, while the Vimba API sets them to 1024. If they were already set to 1024 by Vimba, the camera simply complains but doesn't seam to overwrite its settings, and the image acquisition with Aravis works.

My workaround is to add:
int32_t valx=1024;
arv_device_write_memory (dev,0x000000000002002c,4,&valx,NULL);  //SI_maximum_trailer_size
arv_device_write_memory (dev,0x0000000000020018,4,&valx,NULL);  //SI_maximum_leader_size
The camera still complains because the API tries to set it to a wrong value afterwards anyway, but at least I can use my camera without initializing it with the proprietary API.

Also, I found some unprotected Basler camera on my network, and again the Aravis API could not retrieve any images from it until I added the workaround, after which it worked fine.

Also thanks for the awesome API!

EmmanuelP:
This should be fixed by Johannes in commit 16888e9e94fcf5007f6006e47a0e2696612dab11.
Please reopen if it not the case.

PR #214, opened 2018-12-08
Fix alignment for leader, trailer and payload transfer size.

JohannesBeck:
This PR fixes the alignment for leader, trailer and payload transfer.
The alignment is defined by the device and can be determined using
the SI Info field. The leader and trailer size is aligned according
to the provided value.

To align the payload size the maximum payload size defined by
the host is first aligned by 'rounding' downwards and than the
'SI Payload Transfer Size' is determined. The last step is to
also align the 'SI Final Transfer1 Size'.

### DIFF
--- a/src/arvuvstream.c
+++ b/src/arvuvstream.c
@@ -218,6 +218,10 @@ arv_uv_stream_thread (void *data)
 
 /* ArvUvStream implemenation */
 
+static guint32 align(guint32 val, guint32 alignment) {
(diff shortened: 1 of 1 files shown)


JohannesBeck:
This would probably fix #209.

EmmanuelP:
Hi Johannes,

Thanks a lot for the PR. I will review and test it next week.

Cheers.

EmmanuelP:
Pushed to master with some coding style fixes.

Thanks a lot.

Cheers.

PR #215, opened 2018-12-10
Add nullptr check to uv_device->priv->genicam.

JohannesBeck:
This PR adds a nullptr check to uv_device->priv->genicam. This
pointer can be null if the device is already in use and aravis
cannot connect to the camera. Without this check a
'critical g_object_unref assertion 'g_is_object (object)' failed'
will be asserted.


EmmanuelP:
Don't bother, I'll make the change.

Thanks.

Issue #221, opened 2019-01-12
arv_camera_abort_acquisition leaves unreleased buffer.


wuyuanyi135:
I am memchecking my app with Valgrind. I found that `arv_camera_abort_acquisition` will cause memory leak of an unreleased `ArvBuffer`. I think PR #193 may addressed this problem that the current buffer may not be properly destroyed when finalizing the stream.

My current workaround is to put the allocated buffers in somewhere and check whether any unreleased after the stream gets `g_object_unref`ed.

EmmanuelP:
Fix merged.

PR #223, opened 2019-01-12
fix memory leak: unreleased buffer when acquisition aborts

wuyuanyi135:
#221


EmmanuelP:
Pushed to master with coding style fixes. Thanks.

wuyuanyi135:
@EmmanuelP Same question, does this fix also apply for GigaE?

EmmanuelP:
It is a completely different code, exercized by the test suite (which is not the case for the Usb3Vision ocde).

PR #229, opened 2019-02-05
faster stream destruction

thecaptury:
This commit avoids a 1 second timeout when shutting down a stream by creating a pipe that is used to signal the stream thread. This is particularly useful when working with multiple cameras.

This was the simplest method I could find to terminate g_poll() early.


EmmanuelP:
HI Nils,

Thanks for your work on this issue. It is the right approach, but I would love if you can adapt your code to use the GCancellable API though. Looking at the GCancellable documentation, it seems relatively easy.

thecaptury:
Yes. GCancellable looks nicer than pipe actually.

EmmanuelP:
I have implemented the GCancellable way, in both standard socket and packet socket cases. It is now in master.

Thanks.

PR #234, opened 2019-03-04
Set SO_RCVBUF size to a larger value on discovery socket.

mlrsmith:
This makes discovery work for all the GV devices on networks containing
moderately large numbers of them (in my testing, around 23-24 of them).


EmmanuelP:
Thanks.

PR #239, opened 2019-03-28
Properly interpret "Register" nodes

NathanMOlson:
I came across the need for this when trying to access UserFile1 on a Point Grey Blackfly S.


EmmanuelP:
Thanks.

PR #256, opened 2019-07-05
change order of operations in arv_camera_set_region 

rickvavilin:
For example, if camera has Height==MAX_HEIGHT, we can not set OffsetY>0,
so we need set width and height first, and set offsets then.

### DIFF
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -231,14 +231,14 @@ arv_camera_set_region (ArvCamera *camera, gint x, gint y, gint width, gint heigh
 {
 	g_return_if_fail (ARV_IS_CAMERA (camera));
 
-	if (x >= 0)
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Thanks Aleksandr.

I think your patch is just inverting the issue, but does not fix it.

There should be some logic first trying to find which order has better chance to succeed.

rickvavilin:
Yes, you are right. If width or height values is greater than the current, issue may occur.
Thus, the logic should depend on the current values of the coordinates of the region.

EmmanuelP:
Another solution would be to set x and y to zero first, then apply the region.

rickvavilin:
Set x and y to zero, set width and height, and then set x and y to needed value.

EmmanuelP:
Yes.

EmmanuelP:
Thanks.

PR #263, opened 2019-08-02
Fixed align assert and handle alignment of zero correctly.

JohannesBeck:
For some cameras the SI info register is zero, which not handles correctly. This PR fixes the alignment and the assert.


EmmanuelP:
Thanks Johannes.

gaoyichuan:
Thanks everyone for the quick fix!

PR #265, opened 2019-08-05
Fixes for GVCP:

highperformancecoder:
- ACK id must match discovery id
  - packet_type field may have additional flags indicating broadcast acknowledge
  - Remove discovery_socket, and bind GVCP socket to all interfaces
  - Add Netmask, Mac address, and version to discovery acknowledge

These were required to get the Aravis fake camera discoverable with the Pleora eBUS player. The last item about adding the netmask, mac address etc were possibly not essential, but added because the Player expected that information to be present in the GVCP discovery response packet.

### DIFF
--- a/src/arvgvcp.c
+++ b/src/arvgvcp.c
@@ -348,7 +348,7 @@ arv_gvcp_packet_new_discovery_cmd (size_t *packet_size)
  */
 
 ArvGvcpPacket *
-arv_gvcp_packet_new_discovery_ack (size_t *packet_size)
+arv_gvcp_packet_new_discovery_ack (size_t *packet_size, int id)
 {
 	ArvGvcpPacket *packet;
 
@@ -361,7 +361,7 @@ arv_gvcp_packet_new_discovery_ack (size_t *packet_size)
 	packet->header.packet_type = g_htons (ARV_GVCP_PACKET_TYPE_ACK);
 	packet->header.command = g_htons (ARV_GVCP_COMMAND_DISCOVERY_ACK);
 	packet->header.size = g_htons (ARV_GVBS_DISCOVERY_DATA_SIZE);
-	packet->header.id = g_htons (0xffff);
+        packet->header.id = g_htons (id);
 
 	return packet;
 }
--- a/src/arvgvcp.h
+++ b/src/arvgvcp.h
@@ -266,7 +266,7 @@ ArvGvcpPacket * 	arv_gvcp_packet_new_write_register_cmd 	(guint32 address, guint
 ArvGvcpPacket * 	arv_gvcp_packet_new_write_register_ack 	(guint32 data_index,
 								 guint16 packet_id, size_t *packet_size);
 ArvGvcpPacket * 	arv_gvcp_packet_new_discovery_cmd 	(size_t *size);
-ArvGvcpPacket * 	arv_gvcp_packet_new_discovery_ack 	(size_t *packet_size);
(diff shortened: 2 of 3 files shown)


EmmanuelP:
Hi,

Thanks for your work in theses improvements.

I would prefer if you could split the commit in several ones, each one with a specific change (I think one for each item of your change list would be nice). That is easier to review, and makes 
regression also easier to spot.

Regarding the coding style, please don't declare variables on the fly and don't use C++ comments.

What is the reason for the removal of discovery_socket ?

highperformancecoder:
Sure - I'll have a go at splitting the changes. The trouble is I needed all the changes together in order to test it :) (save possibly the last item on that list).

By "declaring variables on the fly" do you mean the modern practice of declaring a variable as close as possible to its usage, rather than the old fashioned style (eg enforced in Fortran) of declaring all variables at the top of a function before a line of executable code appears? No worries about conforming to style, though.

The reason why I removed discovery socket is that the GVCP listener needed to listen to both the all broadcast address, as well as the subnet broadcast address and the actual IP no of the interface, which is most conveniently achieved by binding to 0.0.0.0 (AKA INADDR_ANY). This conflicted with the additional socket connection to the GVCP port on the named interface. However, I could have minimally done what I needed by adding an additional socket that bound to the subnet broadcast address, which would have required an additional socket member in the private data, as well as code to extract the subnet from the address and form the subnet broadcast address. It's a much more complex solution.

highperformancecoder:
Just checked on variable declarations used within code (what i think you mean by "on the fly") - this was added to the C99 standard, but would not be compliant with earlier versions of C, eg the one described in the last K&R book, which dates back to 1988. Similarly with the // style comment, which was added to C99. It's been so long since I've used "pure C", that I tend to forget these things. I just tried setting -std=c89, but it threw up way to many errors to be useful :(. And it proved surprisingly difficult to get meson to insert the directive into the build.ninja file - I ended up having to do it by hand - story for another time...

highperformancecoder:
OK - I have split the above into 4 separate pull requests. Note that when merging them together, there will be a merge conflict between the 3rd and 4th PR. I can fix the merge conflict later by rebasing the last PR before you merge it in.

Also there is one other change I needed to make, which I won't raise as a PR, since I don't understand why it is needed. I needed to limit the payload size computed on arvgvfakecamera.c:370 to a maximum of 15,000 bytes. 15,000 bytes was determined empirically, but doesn't correspond to anything I could understand - eg if I set the limit to twice a jumbo packet size, eBUSPlayer gets upset.

NB ninja test is reporting a failure on one of the regression tests - it is doing this on the current top of tree, so I don't think I broke anything.

EmmanuelP:
@highperformancecoder 

> Just checked on variable declarations used within code (what i think you mean by "on the fly") - this was added to the C99 standard, but would not be compliant with earlier versions of C, eg the one described in the last K&R book, which dates back to 1988. Similarly with the // style comment, which was added to C99. It's been so long since I've used "pure C", that I tend to forget these things. I just tried setting -std=c89, but it threw up way to many errors to be useful :(. And it proved surprisingly difficult to get meson to insert the directive into the build.ninja file - I ended up having to do it by hand - story for another time...

It is not that I want to follow an ancient standard like C89. The convention of using only /* */ for comments is quite current in the C world, even in the modern one. As well is the declaration of variables at the start of code blocks. The later is the most important to me, as it makes the review for memory leak easier if you have all the variables in a single spot.

PR #266, opened 2019-08-06
ACK id must match discovery id

highperformancecoder:


EmmanuelP:
Pushed to master with a formatting fix and meson change removed

PR #267, opened 2019-08-06
packet_type field may have additional flags indicating broadcast ackn…

highperformancecoder:
…owledge


EmmanuelP:
I have pushed a more complete fix to this issue, by handling separately the two first bytes of the gvcp packet header.

Commit is 48e775674618cb6ceb486b0c366b0c1e5224a5d8

PR #268, opened 2019-08-06
Remove discovery_socket, and bind GVCP socket to all interfaces

highperformancecoder:


src/arvgvfakecamera.c
```diff
@@ -444,43 +444,38 @@ arv_gv_fake_camera_start (ArvGvFakeCamera *gv_fake_camera)
...
 			g_object_unref (inet_socket_address);
 
-			inet_socket_address = g_inet_socket_address_new (inet_address, 0);
-			gv_fake_camera->priv->gvsp_socket = g_socket_new (G_SOCKET_FAMILY_IPV4,
-							       G_SOCKET_TYPE_DATAGRAM,
-							       G_SOCKET_PROTOCOL_UDP, NULL);
-			if (!g_socket_bind (gv_fake_camera->priv->gvsp_socket, inet_socket_address, FALSE, NULL))
-				binding_error = TRUE;
-
 			g_clear_object (&inet_socket_address);
```
EmmanuelP: Double free of inet_socket_address

src/arvgvfakecamera.c
```diff
@@ -444,43 +444,38 @@ arv_gv_fake_camera_start (ArvGvFakeCamera *gv_fake_camera)
...
 			g_clear_object (&socket_address);
 
 			inet_address = g_inet_address_new_from_string ("255.255.255.255");
 			discovery_address_string = g_inet_address_to_string (inet_address);
 			arv_debug_device ("[GvFakeCamera::start] Discovery address = %s", discovery_address_string);
 			inet_socket_address = g_inet_socket_address_new (inet_address, ARV_GVCP_PORT);
```
EmmanuelP: This code is now useless

src/arvgvfakecamera.c
```diff
@@ -444,43 +444,38 @@ arv_gv_fake_camera_start (ArvGvFakeCamera *gv_fake_camera)
...
-				gv_fake_camera->priv->discovery_socket = NULL;
-			else {
-				gv_fake_camera->priv->discovery_socket = g_socket_new (G_SOCKET_FAMILY_IPV4,
-									    G_SOCKET_TYPE_DATAGRAM,
-									    G_SOCKET_PROTOCOL_UDP, NULL);
-				if (!g_socket_bind (gv_fake_camera->priv->discovery_socket, inet_socket_address, FALSE, NULL))
-					binding_error = TRUE;
-				g_socket_set_blocking (gv_fake_camera->priv->discovery_socket, FALSE);
-			}
+                        gv_fake_camera->priv->discovery_socket = NULL;
```
EmmanuelP: priv->discovery_socket is never used, so it should be completely removed

src/arvgvfakecamera.c
highperformancecoder: Yes - I think it was removed in the original pull request, but got missed in the second one.

src/arvgvfakecamera.c
highperformancecoder: Ouch! Something that RAII does naturally for you. There's a reason I haven't done pure C for nearly 30 years :)

highperformancecoder:
No, as it stands, you can only run one camera on a system. If you wanted to run multiple cameras on a system, each attached to a different interface, then you have to add an extra discovery socket. The problem I'm trying to solve is that the GVCP server needs to listen to both the global broadcast address (255.255.255.255) and the subnet broadcast address (eg 10.0.0.255).

Apparently, you can't bind the same socket to multiple addresses, alas. Also, can you bind multiple sockets to the same address and port (in this case 255.255.255.255)? You can't bind multiple sockets to the same port and interface (ie the second bind of 10.0.0.1/3956 fails).

The question to ask is whether you want this functionality?

src/arvgvfakecamera.c
highperformancecoder: Yeah - I had a TODO about that...

highperformancecoder:
I have fixed up the double free and removed the extraneous discovery_socket stuff. I'll also be committing a pull request with two discovery sockets, one on 255.255.255.255 and the other on the subnet broadcast address. This latter version hopefully has the advantage of being able to run multiple cameras on the one machine.

Choose one PR, and delete the other.

EmmanuelP:
I prefer the PR #271 implementation. I'm closing this one.

PR #269, opened 2019-08-06
Add Netmask, Mac address, and version to discovery acknowledge

highperformancecoder:
eBUS Player is not happy without this extra info, so it is needed.


highperformancecoder:
I have removed the netmask, mac address and protocol version from this pull request. The reason being that the eBUS viewer doesn't need it, and that netmask is not even appropriate for IPv6, and mac address only relevant for ethernet. Even though the protocol is called "Gig Ethernet Vision", it is really just another IP protocol, not tied to ethernet :)

EmmanuelP:
The resulting pull request does nothing now, except for setting inet_address twice.

I'm closing it.

PR #272, opened 2019-08-12
Added accessor method for the camera class of ArvGvFakeCamera

highperformancecoder:
I needed this to write a custom gv fake camera program to support the work I'm doing.

### DIFF
--- a/src/arvgvfakecamera.c
+++ b/src/arvgvfakecamera.c
@@ -565,14 +565,15 @@ _set_property (GObject *object, guint prop_id, const GValue *value, GParamSpec *
 }
 
 /** 
-    arv_gv_fake_camera_get_camera
-    Returns: (transfer none): the ArvFakeCamera instance of this ArvGvFakeCamera object
+ *  arv_gv_fake_camera_get_camera:
+ *  Returns: (transfer none): the #ArvFakeCamera instance of this #ArvGvFakeCamera object
  */
-ArvFakeCamera* arv_gv_fake_camera_get_camera(ArvGvFakeCamera *gv_fake_camera)
+
+ArvFakeCamera *
+arv_gv_fake_camera_get_camera (ArvGvFakeCamera *gv_fake_camera)
 {
         return gv_fake_camera->priv->camera;
 }
-
 
 static void
 arv_gv_fake_camera_init (ArvGvFakeCamera *gv_fake_camera)
--- a/src/arvgvfakecamera.c
+++ b/src/arvgvfakecamera.c
@@ -564,6 +564,10 @@ _set_property (GObject *object, guint prop_id, const GValue *value, GParamSpec *
 	}
 }
 
+ArvFakeCamera* arv_gv_fake_camera_get_camera(ArvGvFakeCamera *gv_fake_camera)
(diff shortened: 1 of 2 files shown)


src/arvgvfakecamera.c
```diff
@@ -564,6 +564,10 @@ _set_property (GObject *object, guint prop_id, const GValue *value, GParamSpec *
 	}
 }
 
+ArvFakeCamera* arv_gv_fake_camera_get_camera(ArvGvFakeCamera *gv_fake_camera)
+{return gv_fake_camera->priv->camera;}
```
EmmanuelP: Coding style issue

src/arvgvfakecamera.h
```diff
@@ -57,6 +57,12 @@ ArvGvFakeCamera *		arv_gv_fake_camera_new		(const char *interface_name);
 gboolean 			arv_gv_fake_camera_start	(ArvGvFakeCamera *gv_fake_camera);
 void	 			arv_gv_fake_camera_stop		(ArvGvFakeCamera *gv_fake_camera);
 
+/** 
+    arv_gv_fake_camera_get_camera
+    Returns: (transfer none): the ArvFakeCamera instance of this ArvGvFakeCamera object
```
EmmanuelP: API documentation should go into c files

src/arvgvfakecamera.c
highperformancecoder: I guess you mean you want the braces on their own lines...

src/arvgvfakecamera.h
highperformancecoder: That genuinely seems arse about. .h files are the interface files, and that's where you'd naturally go to to see documentation on how to use stuff, as opposed to documentation about implementation detail.
But hey, it's your project.

EmmanuelP: It's better if the documentation is near the implementation. Make it easier to not forget to change it when your working on the code. Also, a change in a c file only triggers the recompilation of one file, while a modification of a header file may cause the recompilation of the whole project.

src/arvgvfakecamera.c
EmmanuelP: Yes. The return value should be on its own line. Please also take care of the proper spacing.

highperformancecoder: Trouble is that that would be fighting my editor, not to mention muscle memory. My editor needs to be set to 2 space indenting for other projects. If its so important, there are pretty formatting tools that can be used to enforce a particular spacing... Personally, I don't think it is too important.

src/arvgvfakecamera.h
highperformancecoder: When working on an implementation, one shouldn't be changing the semantics of the interface. Some might go as far as saying one should right the documentation before implementing it - I'm not so extreme, of course. 

More to the point, code is read more often than written - and consulting the interface file is where you go for advice, not the implementation file. I actually hadn't noticed the documentation much at all until you mentioned it.

Re compilation times - its been years since I've worked on a project where compilation time was an issue. I don't think Aravis is in any danger of having this problem!

:)

PR #273, opened 2019-08-12
Fix payload size calculations for pixel sizes > 1 byte

highperformancecoder:
This fixes a problem with buffer sizes being incorrect when multibyte pixels are used, eg the colour RGB8 format, or MONO16 format.

I'm not quite sure my fix for the Camera class is correct, as probably "PayloadSize" should be computed in full on the camera device, and transmitted over the wire, however I don't know how I can represent the bits per pixel calculation in Swiss Knife.

Nevertheless, the code works fine with eBUSPlayer, probably because the latter does not make use of PayloadSize, but rather computes it directly from width, height and pixel format.


src/arvfakecamera.h
```diff
@@ -134,6 +134,11 @@ void		arv_fake_camera_set_fill_pattern	(ArvFakeCamera *camera,
 							 void *fill_pattern_data);
 void 		arv_fake_camera_set_trigger_frequency 	(ArvFakeCamera *camera, double frequency);
 
+/**
```
EmmanuelP: I welcome documentation changes, but that should be in another commit.

Please refer to the gtk-doc documentation: https://developer.gnome.org/gtk-doc-manual/unstable/documenting_syntax.html.en

EmmanuelP:
As a bonus point, it would be nice to have arv_fake_camera_diagonal_ramp generate a pattern for every supported pixel formats.

src/arvfakecamera.h
highperformancecoder: Sorry - yes that should not be here. And as discussed in a previous PR, it needs to go in the .c file :).

src/arv-fake-camera.xml
```diff
@@ -222,6 +222,9 @@
 		<EnumEntry Name="Mono8" NameSpace="Standard">
 			<Value>17301505</Value>
 		</EnumEntry>
+		<EnumEntry Name="RGB8" NameSpace="Standard">
+			<Value>35127316</Value>
+		</EnumEntry>
```
highperformancecoder: Actually, this edit slipped in too. I was trying to figure out how get different pixel formats, and ultimately realised what was needed was a custom genicam.xml file, which led me down to puzzling over the use of arv_set_fake_camera_genicam_filename.

highperformancecoder:
> As you said, the fix is incomplete. The PayloadSize register must be computed correctly for every available PixelFormat.
> 
> Here is what is done by the TheImagingSource cameras:
> 
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml:  <IntSwissKnife Name="CalcBytesPerPixel" NameSpace="Custom">
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml-    <pVariable Name="PIXELFORMAT">PixelFormatRegister</pVariable>
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml-    <Formula>((PIXELFORMAT &gt;&gt; 16) &amp; 0xFF) / 8</Formula>
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml-  </IntSwissKnife>
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml-  <IntSwissKnife Name="CalcPayloadSize" NameSpace="Custom">
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml-    <pVariable Name="WIDTH">Width</pVariable>
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml-    <pVariable Name="HEIGHT">Height</pVariable>
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml:    <pVariable Name="BYPP">CalcBytesPerPixel</pVariable>
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml-    <Formula>WIDTH * HEIGHT * BYPP</Formula>
> TheImagingSource-DMK-23G618-1.0.0-GigEVision.xml-  </IntSwissKnife>

Thanks - this looks like the ticket!

EmmanuelP:
Pushed with some minor fixes. Thanks.

PR #276, opened 2019-08-14
global discovery socket needs to be marked SO_REUSEPORT to allow

highperformancecoder:
multiple fake cameras on the same system.

I have just marked the global discovery socket as reusable, which allows multiple cameras provided they are bound to different interfaces. IIUC correctly, Linux will shard the incoming UDP messages across processes - I tested this with arv_viewer, arv-fake-camera and another camera simulator I have written for my own purposes, and that works. Two instances of arv-fake-camera do not show up in the viewer because they both have the same serial number (GV01). Perhaps we can add the interface name somehow into the mix to distinguish them.


EmmanuelP:
Thanks.

PR #284, opened 2019-08-30
Clarify scale of the timeout variable.

abrock:
Just a minor clarification, I was assuming the timeout is in seconds, fortunately I checked before using a value of 1 :-)


EmmanuelP:
Thanks.

PR #296, opened 2019-09-25
viewer: Store appstream file in metainfo dir

rffontenelle:
"appdata" is the legacy path for AppStream files, and now such directory should be called "metainfo".

https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps


EmmanuelP:
Thanks.

PR #305, opened 2019-10-22
uvdevice: reset data endpoint in _open_usb_device

### LINKED ISSUE #301: Problem while replugging the camera
**Describe the bug**
I have application that operates two basler USB cameras. This application handles re-plugging of the cameras. Most of the times when I re-plug the camera everything goes smoothly: aravis is being able to claim USB interface and operate the camera (configure the camera and gather images). But sometimes re-plugging of one of the cameras causes *another* camera to be in a broken state.
I can access cameras in this state in a sense that I can query genicam data and different feature values, but it fails to return images, emitting various "USB transfer error" - LIBUSB_ERROR_TIMEOUT or LIBUSB_ERROR_IO

**To Reproduce**
Steps to reproduce the behavior: Unfortunately I cannot reproduce this issue reliably even with my own application. But basically it is: have two usb cameras running and acquiring images, unplug and re-plug one of the cameras, sometimes this causes *another* camera to enter in a broken state in which they fail to get any images.

**Expected behavior**
I expected that re-plugging of one of the cameras should not affect the other camera.

**Camera description:**
 - Manufacturer 2x Basler
 - Model 2x acA4096-30uc
 - Interface 2x USB3

**Platform description:**
 - Aravis version 0.6.2
 - OS: Ubuntu 18.04
 - Hardware x86_64

**Additional context**
Here is the log of ARV_DEBUG=all:6 when I try to start cameras in a broken state.  [broken-state.log](https://github.com/AravisProject/aravis/files/3708006/broken-state.log)
I also have a larger log where this issue is triggered with libusb debug information, but I need to find a way to clean-up this log a bit since it is too big to upload and to read.

(Powercycling of the cameras of course helps, but I was hoping to avoid that.)

I didn't managed to pinpoint exact reason for that type of behavior but I have just a minor concern that might or might not contribute to this. Sometimes *during* the re-plug I'm getting LIBUSB_ERROR_PIPE error, and it seems like this kind of error requires a call to libusb_clear_halt before retrying libusb_bulk_transfer (as it is described [here](https://devzone.nordicsemi.com/f/nordic-q-a/41390/usb-endpoint-transfer-pipe-error))
>Looking at the documenation here, libusb_bulk_transfer () returning LIBUSB_ERROR_PIPE , means that the endpoint is halted. I think you might need to call libusb_clear_halt, and then try again.

crabsmaps:
Fixes #301 

With this PR I no longer have problems with replugging the camera.

I think that the main reason for that problem was that the camera has been in a halted state (see Figure 5-3:  Transmission States in [USB3 Vision Specification](https://www.visiononline.org/userAssets/aiaUploads/file/USB3_Vision_Specification_v1-0-1.pdf))

In order to return camera back to the idle state I'm setting data endpoint in a halted state and immediately clearing that state (this solution I picked up from [here](https://github.com/ni/usb3vision/blob/master/u3v_core.c#L1488) and from [here](https://github.com/signal11/m-stack/blob/master/host_test/feature.c#L97))

### DIFF
--- a/src/arvuvdevice.c
+++ b/src/arvuvdevice.c
@@ -620,6 +620,29 @@ arv_uv_device_get_genicam_xml (ArvDevice *device, size_t *size)
 }
 
 static void
+reset_endpoint (libusb_device_handle *usb_device, guint8 endpoint, guint8 endpoint_flags)
+{
+		// Set endpoint in halt condition
+		int errcode = libusb_control_transfer(usb_device,
+			LIBUSB_ENDPOINT_OUT|LIBUSB_REQUEST_TYPE_STANDARD|LIBUSB_RECIPIENT_ENDPOINT,
+			LIBUSB_REQUEST_SET_FEATURE,
+			0, /*wValue: 0=endpoint_halt*/
+			endpoint|endpoint_flags,
+			0, 0,
+			1000);
+		if (errcode < 0) {
+			arv_warning_device("Failed to set endpoint %x in halt condition: %s", endpoint|endpoint_flags, libusb_error_name(errcode));
+			return;
+		}
+		// Clear halt condtion on the endpoint, effectivelly resetting the pipe
+		errcode = libusb_clear_halt(usb_device, endpoint | endpoint_flags);
+		if (errcode < 0) {
+			arv_warning_device("Failed to clear halt contidion on endpoint: %s", libusb_error_name(errcode));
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Thanks Dmitry.

I have pushed your fix to master (with some minor cosmetic fixes).

Cheers,

    Emmanuel.

PR #306, opened 2019-10-26
Fix annotation for output parameters.

de-vri-es:
This PR adds the `(out)` annotation to output parameters that was missing in a few places. Where I think I understood the function, I also tried to document them.

The PR also labels the return value of `arv_gc_enumeration_get_available_int_values` as an array with full ownership transfer to the caller.

Without the proper annotations, automatically generating bindings may fail or generate incorrect bindings. I may find more missing annotations when I start using the generated bindings. If so, I will submit more PRs :)


de-vri-es:
Also pushed a commit that adds types to the defines for pixel format constants. Without it, they're listed in the GIR file as `gint`. That wouldn't always be a big poblem, but in this case the constants overflow the range of 32 bit signed integers.

EmmanuelP:
Thanks a lot Maarten.

We have tried to fix the define type issue lately, without success. Normally the type annotation should have fixed that, but it did not work as expected, hence this issue report: https://gitlab.gnome.org/GNOME/gobject-introspection/issues/315

I have pushed your commits to master, with a minor fix in the commit titles.

de-vri-es:
Awesome. Thanks for the quick review and fix :)

PR #317, opened 2019-11-27
Add support for lockable buffers

jgressmann:
Lockable buffers, together with preallocated memory enable grabbing
into a ring buffer with concurrent access by aravis (writer) and its
clients (readers).

### DIFF
--- a/src/arvbuffer.c
+++ b/src/arvbuffer.c
@@ -34,17 +34,6 @@
 
 #include <arvbufferprivate.h>
 
-static gboolean nop_try_lock (ArvBuffer *buffer)
-{
-	(void)buffer;
-	return TRUE;
-}
-
-static void nop_unlock (ArvBuffer *buffer)
-{
-	(void)buffer;
-}
-
 gboolean
 arv_buffer_payload_type_has_chunks (ArvBufferPayloadType payload_type)
 {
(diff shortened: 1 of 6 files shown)


EmmanuelP:
Hi Jean,

Thanks for the pull request. I'm not sure to understand the use case for this API. In a normal use, a buffer is never shared between the frame receiving thread and the clients.

Could you explain how this API is supposed to be used ?

jgressmann:
Thank you for taking the time to look at my pr.

While I was writing the explanation I realized I had (badly) implemented ring buffers on top of ArvBuffer. I have thus updated the code to better reflect what the code is trying to do.

Let's say I have a shared memory segment for images that I want to fill from aravis. This shared memory segment is being read by multiple client processes. Using the conventional approach I would create a set of buffers, add them to the stream and then _copy_ the image into the shared memory segment. This works just fine. The rub is in the copy. It slows things down. 

The change I propose avoids the copy. One sets up preallocated buffers pointing into the shared memory segment and adds these the the camera stream. When a frame is ready, aravis takes the first buffer and places the image. So far, so good. Now problem here is that you want to make sure you are not grabbing into a buffer that is currently being read from. That's where the locking comes into play. Each client will read-lock a buffer it is currently using. When aravis needs to use a buffer, it acquires its write lock, stores the image and then then releases the lock, making it available again to the clients.

EmmanuelP:
With what is currently in aravis, no copy is needed.

You create a set of buffers using the preallocated memory, and push them to the stream input queue. Once a buffer is filled, you pop it from the stream output queue, do whatever you want with it, and when you are done, you push it in the stream input queue. When a buffer is neither in the input nor the output queue, aravis has no access to its content.

jgressmann:
In my use case that would mean I need to write-lock any buffer I push into aravis' input queue to prevent my clients from accessing it. Even if that buffer isn't used right away (or ever). 

Any buffer that I pull out of the output queue I would write-unlock and make available to the clients.  But how to know when I can put it back? There are several client processes running in parallel to the one acquiring the images (img-acq). I can't try to lock the write-lock the buffer in another thread of img-acq. For one that would be a waste of resources and second introduce a race condition between a client trying to read same buffer and the thread trying to return it to aravis. What I am trying to say is that I don't have any way of knowing when all clients are done with a particular buffer. 

Now if I turn things around, that is only write lock the buffer in aravis for the duration I know it is in use, any client can always access all buffer except for the single write-locked one.

EmmanuelP:
`GAsyncQueue` is designed for passing data around threads, and this is what is used by ArvStream.

In your case, you just have to make every processing threads wait for completed buffers using `arv_stream_try_pop_buffer()` or `arv_stream_timeout_pop_buffer()`, process the data, then push back the buffer to ArvStream using 'arv_stream_push_buffer()' when they are done.

jgressmann:
I cannot use GAsyncQueue b/c the image readers (consumers) live in
different os processes than the producer (img-acq). Thats why I am using a
shared memory segment.

EmmanuelP:
If the processing happen in another process, a possible implementation would be:

```
while (not_stop)
  buffer = arv_stream_try_pop_buffer
  for each pending_buffer in to_be_processed_buffer_list {
     if pending_buffer.processing_done {
        arv_stream_push_buffer (pending_buffer)
        remove pending_buffer from to_be_processed_buffer_list
    }
  }
  buffer.processing_done = false
  buffer.ready_to_process = true
  append buffer to to_be_processed_buffer_list
}
```

I appreciate your work in this pull request, but I'd like to keep the complexity of your particular use out case of aravis.

jgressmann:
Looks like I found a way to setup the ring buffer w/o a the need to change anything in aravis.
Thanks again for taking the time to help me out.

PR #329, opened 2020-01-04
print source location details in log messages

makelinux:
It is difficult to work with logs without precises source location, so this solves the issue.


EmmanuelP:
Hi,

Sorry for the delay, but it comes from my reluctance to refuse a pull request on the basis of mostly non objective reasons.

I don't like how the debug output is polluted by what I find mostly noise, useless for understanding what is going on. It is not that hard to find the source of a given debug line, by grepping the source files.

So sorry, I will not accept this pull request.

Thanks anyway,

  Cheers.

PR #343, opened 2020-04-13
meson: fix issues when used as a subproject

rhssk:
- Add missing include dir to the declared dependency
- Clean up required libraries in pkgconfig generation since meson is  smart enough to do that for us.
  .pc diff (meson 0.54.0):
```diff
8,10c8,11
< Requires: glib-2.0 gobject-2.0 gio-2.0
< Requires.private: glib-2.0 >= 2.44, gobject-2.0, gio-2.0, libxml-2.0, zlib, libusb-1.0, libxml-2.0 zlib libusb-1.0
< Libs: -L${libdir} -laravis-0.8 -lm
---
> Requires: glib-2.0 >= 2.44, gobject-2.0, gio-2.0
> Requires.private: libxml-2.0, zlib, libusb-1.0
> Libs: -L${libdir} -laravis-0.8
> Libs.private: -lm
```
  This also fixes a deprecation warning:
```
subprojects/aravis/src/meson.build:256: DEPRECATION: Library aravis-0.8 was passed to the "libraries" keyword argument of a previous call to generate() method instead of first positional argument. Adding aravis-0.8 to "Requires" field, but this is a deprecated behaviour that will change in a future version of Meson. Please report the issue if this warning cannot be avoided in your case.
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -25,14 +25,12 @@ libz_dep = dependency ('zlib', required: true)
 
 libm = cc.find_library ('m', required: true)
 
-aravis_dependencies = [glib_dep, gobject_dep, gio_dep, xml2_dep, libz_dep, libm]
-aravis_dependency_str = 'glib-2.0 gobject-2.0 gio-2.0'
-aravis_private_dependency_str = 'libxml-2.0 zlib'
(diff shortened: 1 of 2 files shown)


EmmanuelP:
Thanks Rihards.

PR #344, opened 2020-04-16
all: fix issues reported by Clang static analyzer

rhssk:
Most of the reported issues are false-positives. The remaining ones
consist of dead code, uninitialized values and one memory leak.

### DIFF
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -1130,6 +1130,11 @@ arv_camera_get_frame_rate_bounds (ArvCamera *camera, double *min, double *max, G
 					return;
 				}
 
+				if (max != NULL)
+					*max = 0;
+				if (min != NULL)
+					*min = 0;
+
 				for (i = 0; i < n_values; i++) {
 					if (values[i] > 0) {
 						double s;
--- a/src/arvgvfakecamera.c
+++ b/src/arvgvfakecamera.c
@@ -265,13 +265,11 @@ _thread (void *user_data)
 
 	do {
 		guint64 next_timestamp_us;
-		guint64 sleep_time_us;
 
 		if (is_streaming) {
-			sleep_time_us = arv_fake_camera_get_sleep_time_for_next_frame (gv_fake_camera->priv->camera, &next_timestamp_us);
+			arv_fake_camera_get_sleep_time_for_next_frame (gv_fake_camera->priv->camera, &next_timestamp_us);
 		} else {
-			sleep_time_us = 100000;
-			next_timestamp_us = g_get_real_time () + sleep_time_us;
+			next_timestamp_us = g_get_real_time () + 100000;
 		}
 
 		do {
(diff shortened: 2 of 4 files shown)


src/arvcamera.c
```diff
@@ -1130,6 +1130,11 @@ arv_camera_get_frame_rate_bounds (ArvCamera *camera, double *min, double *max, G
 					return;
 				}
 
+				if (max != NULL)
```
EmmanuelP, line 1133: If n_values <= 0, we still want to assign a value to min and max. What was the clang static analyzer warning ?

src/arvgvfakecamera.c
```diff
@@ -265,13 +265,11 @@ _thread (void *user_data)
 
 	do {
 		guint64 next_timestamp_us;
-		guint64 sleep_time_us;
 
 		if (is_streaming) {
-			sleep_time_us = arv_fake_camera_get_sleep_time_for_next_frame (gv_fake_camera->priv->camera, &next_timestamp_us);
+			arv_fake_camera_get_sleep_time_for_next_frame (gv_fake_camera->priv->camera, &next_timestamp_us);
```
EmmanuelP, line 270: Ok.

src/arvmisc.c
```diff
@@ -345,11 +345,12 @@ arv_value_copy (ArvValue *to, const ArvValue *from)
 ArvValue *
 arv_value_duplicate (const ArvValue *from)
 {
-	ArvValue *value = g_new (ArvValue, 1);
+	ArvValue *value;
```
EmmanuelP, line 348: Ok.

src/arvstream.c
```diff
@@ -480,8 +480,6 @@ arv_stream_init (ArvStream *stream)
 {
 	ArvStreamPrivate *priv = arv_stream_get_instance_private (stream);
```
EmmanuelP, line 481: Ok.

src/arvcamera.c
rhssk: The warning was about uninitiated values.
E.g. when [`arv_camera_get_frame_rate_bounds`](https://github.com/AravisProject/aravis/blob/master/src/arvcamera.c#L1110) gets called from [`arv_camera_set_frame_rate`](https://github.com/AravisProject/aravis/blob/master/src/arvcamera.c#L941), [these 'greater than' comparisons](https://github.com/AravisProject/aravis/blob/master/src/arvcamera.c#L1139) can be wrong.

`guint n_values` can only be >= 0. In that case:
- if `n_values` == 0: final values of `min` and `max` are 0 (maybe it would be better to set an error?)
- if `n_values` > 0: final values of `min` and `max` are > 0.

rhssk:
> Do you think the static analysis could be automated and integrated in the test target ?

Sure. I used ninja's [`scan-build`](https://mesonbuild.com/howtox.html#use-clang-static-analyzer) target. As mentioned previously it did have quite a bit of false positives in GLib stuff (`g_propagate_error` etc.). I'll check if it supports whitelisting of some sorts.

src/arvcamera.c
EmmanuelP: I'm so stupid, I thought the patch were removing those lines, not adding them.

PR #352, opened 2020-05-05
build: allow to build arv-viewer without optional libusb

lpea:
Despite the fact that libusb is an optional dependency, it looks like the viewer is not built if libusb is not found. The proposed solution is to add usb to `aravis_dependencies` only if it was found. I'm not familiar with meson so it may not be the most elegant solution though.


EmmanuelP:
I can not think of a better solution. Thanks for the patch.

Cheers.

PR #356, opened 2020-05-10
gcpropertynode: don't try to get private member addresses of NULL pointers

rhssk:
When compiled with [AddressSanitizer](https://mesonbuild.com/howtox.html#use-address-sanitizer) a runtime error is detected:
```
../subprojects/Aravis/src/arvgcpropertynode.c:59:1: runtime error: pointer index expression with base 0x000000000000 overflowed to 0xffffffffffffff90
```

This is caused by glib applying negative offset to object pointer in order to get the private member. If object pointer is `NULL` an overflow occurs.
It seems like in all cases the overflow doesn't cause any damage (EDIT: never mind, looks like it's UB) since pointer is checked against `NULL` later on. Nevertheless it would probably be for the best to avoid such cases at all.

There are quite a bit more cases like this in other sources. Fixing all of them would require more time so I only focused on the most glaring one.

### DIFF
--- a/src/arvgcpropertynode.c
+++ b/src/arvgcpropertynode.c
@@ -293,9 +293,9 @@ _get_pvalue_node (ArvGcPropertyNode *property_node)
 const char *
 arv_gc_property_node_get_name (ArvGcPropertyNode *node)
 {
-	ArvGcPropertyNodePrivate *priv = arv_gc_property_node_get_instance_private (node);
-
 	g_return_val_if_fail (ARV_IS_GC_PROPERTY_NODE (node), NULL);
+
+	ArvGcPropertyNodePrivate *priv = arv_gc_property_node_get_instance_private (node);
 
 	return priv->name;
 }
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Hi Rihards,

Thanks for looking into improving aravis code correctness.

The problem is the issue reported by AddressSanitizer is a false positive.
Is it possible to write a suppression rule ?

rhssk:
Hi,

AFAIK arithmetic operations on `NULL` are undefined behavior.
Yes, it's possible to add suppressions to ASAN but I don't think it's the right thing to do in this case...
Why do you feel this is a false positive?

EmmanuelP:
There arithmetic is may be meaningless, but the calculated pointer use is prevented by the NULL check.

EmmanuelP:
Anyway, the pull request still has to be corrected in order to always have variable declaration at the beginning of a code block.

```
const char *
arv_gc_property_node_get_name (ArvGcPropertyNode *node)
{
	ArvGcPropertyNodePrivate *priv;

	g_return_val_if_fail (ARV_IS_GC_PROPERTY_NODE (node), NULL);

	priv = arv_gc_property_node_get_instance_private (node);

	return priv->name;
}
```

kohtala:
I'm not sure if this really needs to be fixed, but if it is it would be interesting if http://coccinelle.lip6.fr/ could apply a patch all over avoiding lots of manual work.

EmmanuelP:
In the end I will not apply this merge request.

The move of the variable assignation introduce a coding style incoherence, and is motivated by a false positive of the sanitizer.

Thanks anyway for your work.

Cheers.

PR #358, opened 2020-05-11
gst: Add packet-delay property

kohtala:
The packet delay is reset on SVS-VISTEK eco424CVGE and exo304CGETR cameras
when GevCCP access is released. Must set it while holding the control
access.

This adds a property as the features property would be more difficult due to the scaling by GevTimestampTickFrequency.


EmmanuelP:
Thanks a lot Marko.

PR #359, opened 2020-05-11
gst: Set pixel format before binning and region

kohtala:
SVS-Vistek cameras reset Width and Height when PixelFormat is set. With the old initialization order the region was always max size regardless of width and height in caps.

### DIFF
--- a/gst/gstaravis.c
+++ b/gst/gstaravis.c
@@ -179,9 +179,9 @@ gst_aravis_set_caps (GstBaseSrc *src, GstCaps *caps)
 
 	pixel_format = arv_pixel_format_from_gst_caps (gst_structure_get_name (structure), format_string, bpp, depth);
 
-	arv_camera_set_region (gst_aravis->camera, gst_aravis->offset_x, gst_aravis->offset_y, width, height, NULL);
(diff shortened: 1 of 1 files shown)


EmmanuelP:
My fear is for some other devices the right order might be the current one. But it this order seems the more logical. Thanks.

kohtala:
> My fear is for some other devices the right order might be the current one. But it this order seems the more logical. Thanks.

My thought exactly. There are quite a number of features that could invalidate others in different ways compared to other devices. It feels like a mine-field.

I've been thinking it would need some sort of container where you can place the values you want. It would then order them according to the genicam definition and could read after write to verify they all got set as intended.

The container could also contain values we'd like to know. The protocol allows reading more than one register in one request so that would allow batch reading. The container would then also be usable as a cache of known device state to help updates without round-robin to camera. If the camera supports reporting of changed values, the container could implement GObject signals to notify of changes. In any case, if one part of program changed some value, it could report that.

It also seems some features are not implemented on all devices the same, like the framerate. The genicam definition could then be patched to add missing features using those that are there. That way all the dependencies between features would be preserved.

But all that is a bit too much for me. I have a practical need, but time is short.

This seemed like a practical solution.

PR #360, opened 2020-05-11
gst: Fix auto gain, exposure and binning on init

### LINKED ISSUE #357: aravissrc gain-auto property has no effect on plugin init
**Describe the bug**
Camera always receives GainAuto=Continuous regardless of gain-auto parameter.

**To Reproduce**
Eg:

    gst-launch-1.0 aravissrc gain-auto=false ! video/x-raw,width=128,height=64,framerate=1/1 ! fakesink

**Expected behavior**
gain-auto=false sets GainAuto=Off.

**Camera description:**
 - SVS-Vistek or Basler or any
 - Any
 - GigE

**Platform description:**
 - Aravis version ARAVIS_0_7_4-26-gb537efc
 - OS: debian:bullseye-20200422
 - Hardware x86_64

**Additional context**

gst_aravis_init (GstAravis *gst_aravis) has

	gst_aravis->gain_auto = FALSE;

gst_aravis_set_property is called when the instance is just made does

		case PROP_GAIN_AUTO:
			gst_aravis->gain_auto = g_value_get_boolean (value);
			if (gst_aravis->camera != NULL)
				arv_camera_set_gain_auto (gst_aravis->camera, gst_aravis->gain_auto, NULL);
			break;

gst_aravis_start calls gst_aravis_init_camera to overwrite these when camera is started

	gst_aravis->gain = arv_camera_get_gain(gst_aravis->camera, &error);
	if (!error) gst_aravis->gain_auto = arv_camera_is_gain_available(gst_aravis->camera, &error);

and later gst_aravis_set_caps does

	if(gst_aravis->gain_auto) {
		arv_camera_set_gain_auto (gst_aravis->camera, ARV_AUTO_CONTINUOUS, NULL);
		GST_DEBUG_OBJECT (gst_aravis, "Auto Gain = continuous");

The property has no effect and gain_auto is set to continuous wheneve the device has Gain.

There seems to be other properties also set in gst_aravis_init_camera may be broken also.

I can make a pull request.

kohtala:
Reading properties from camera in gst_aravis_init_camera overwrites
properties set on launch string.

Fixes #357

### DIFF
--- a/gst/gstaravis.c
+++ b/gst/gstaravis.c
@@ -211,25 +211,29 @@ gst_aravis_set_caps (GstBaseSrc *src, GstCaps *caps)
 
 	GST_DEBUG_OBJECT (gst_aravis, "Actual frame rate = %g Hz", arv_camera_get_frame_rate (gst_aravis->camera, NULL));
 
-	if(gst_aravis->gain_auto) {
-		arv_camera_set_gain_auto (gst_aravis->camera, ARV_AUTO_CONTINUOUS, NULL);
-		GST_DEBUG_OBJECT (gst_aravis, "Auto Gain = continuous");
-	} else {
+	if (gst_aravis->gain_auto_set) {
+		arv_camera_set_gain_auto (gst_aravis->camera, gst_aravis->gain_auto, NULL);
+		GST_DEBUG_OBJECT (gst_aravis, "Auto Gain = %s", arv_auto_to_string(gst_aravis->gain_auto));
+	}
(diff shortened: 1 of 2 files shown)


EmmanuelP:
Thanks.

I would not mind another pull request for the conversion of the auto properties to enums :)

The API can be broken, that is allowed between two stable series.

kohtala:
I'll see if I can come up with something for the enum. I have no use for it myself, so difficult to make a promise as other tasks could eat the time.

PR #362, opened 2020-05-15
Update arvgcregisternode.c

mrc-sys:
Some registers are write only "WO" registers e.g GevTimestampControlLatch. Therefore evaluate the register node's access_mode and read only if the node is not "WO".


EmmanuelP:
Thanks.

PR #368, opened 2020-05-19
Fix spelling available, Unknown, endianness

kohtala:
Debian linter highlighted some common misspellings in the binaries.
"endianess" still reads in the API, but this does not change API.


EmmanuelP:
Thanks.

PR #370, opened 2020-06-01
gst: Rename num-buffers to arv-buffers

kohtala:
GstBaseSrc implements num-buffers property already and this prevents that functionality. Using own name we get to limit the number of frames captured with num-buffers, as expected.


kohtala:
> I'm ok with this change, but could you use num-arv-buffers instead of just arv-buffers ?

Sure. I'll update the patch.

EmmanuelP:
Thanks.

PR #379, opened 2020-07-04
Implements arv_gc_feature_node_get_name_space() function.

sh0:
This function is useful for retrieving feature node name space value.

In some cases camera manufacturers do not give detailed display text or description text for some standard feature nodes. By checking feature name and also if standard name space is used in particular feature, we can add our own display text. This is the basic use case.


EmmanuelP:
Thanks for the patch.

Cheers.

PR #381, opened 2020-07-07
arvtool: png snapshot

makelinux:


src/arvtool.c
```diff
@@ -190,6 +193,43 @@ arv_tool_list_features (ArvGc *genicam, const char *feature, ArvToolListMode lis
...
+	} else {
+		g_autoptr(GDateTime) time = g_date_time_new_now_local();
+		// sortable, compact 12 digit and still readable timestamp
+		fn = g_date_time_format(time, "A-%y%m%d+%H%M%S.png");
+	}
+
+	const char *data;
+	data = arv_buffer_get_data (buffer, NULL);
+
+	png_image image = { .version = PNG_IMAGE_VERSION, .format= PNG_FORMAT_GRAY,
```
EmmanuelP, line 227: PNG_FORMAT_GRAY is used here while the buffer may contain another pixel format.

src/arvtool.c
```diff
@@ -410,6 +450,12 @@ main (int argc, char **argv)
 
 	arv_debug_enable (arv_option_debug_domains);
 
+	if (argc > 1 && g_strcmp0 (argv[1], "snapshot") == 0) {
+		snapshot(argc, argv);
```
EmmanuelP, line 454: snapshot should be integrated in arv_tool_execute_command.

src/meson.build
```diff
@@ -269,7 +269,10 @@ foreach p: progs
 		p[1],
 		link_with: aravis_library,
 		c_args: library_c_args,
-		dependencies: aravis_dependencies,
+		dependencies: aravis_dependencies +
```
EmmanuelP, line 272: Dependency on libpng should be optional.

EmmanuelP:
Implemented using gstreamer.

PR #383, opened 2020-07-08
Adds number of arv_gc_register_description_node_get_* functions.

sh0:
These functions are useful for querying various RegisterDescription node attributes.

I have an application that allows exploring GenICam-compliant camera properties. It would be useful to also show current GenICam XML version. Right now I would need to parse the raw XML to find these parameters.


EmmanuelP:
Thanks.

PR #384, opened 2020-07-09
Better string register writing support.

sh0:
Before this patch all strings were written without null-terminator to the device. This may cause issues when writing over longer strings in camera memory and fails completely when writing zero-length strings as it would amount to writing zero bytes to camera registers.

New string writing method first obtains maximum allowed string length. If the string length is more than allowed then the function returns with an error. Otherwise the string is written to the memory with null-terminator as long as there is enough space. If string length is equal to maximum allowed length then null-terminator is omitted.


EmmanuelP:
Thanks.

PR #386, opened 2020-07-13
usb3vision: Add Daheng Imaging USB rules

Wallbraker:


EmmanuelP:
Thanks Jakob for the patch.

PR #387, opened 2020-07-17
genicam: add documentation related to ArvGcNameSpace

sh0:
The `ArvGcNameSpace` enum and respective feature node function was added in PR #379. However, the documentation was missing which is now added in this patch.


EmmanuelP:
Thanks.

PR #388, opened 2020-07-17
genicam: add 'Representation' and 'Unit' property node support for swiss knife

sh0:
This does not change public API, but interface functions are now able to get SwissKnife and IntSwissKnife Unit and Representation node info.


EmmanuelP:
Thanks.

Issue #332, opened 2020-01-10
Complete arv_gc_feature_node_get_access_mode implementation


madamczyk7ce:
As is written in the topic of issue and here:
https://aravis-project.discourse.group/t/loading-genicam-nodes-to-camera/62/7

sh0:
@EmmanuelP I am interested in completing `arv_gc_feature_node_get_access_mode` functionality so I did some digging.

As far as I understand, feature nodes should only have `ImposedAccessMode` property and registers should only have `AccessMode` property. Right now feature node implements both `ImposedAccessMode` and `AccessMode`. Is there a reason for this? Maybe some camera manufacturer uses it by mistake?

Current idea is to check referenced nodes using `arv_gc_property_node_get_access_mode` and then impose further restrictions using `ImposedAccessMode` if it is available. So should I also keep `AccessMode` for feature node and check it in addition to `ImposedAccessMode`?

EmmanuelP:
Hi Siim,

> @EmmanuelP I am interested in completing arv_gc_feature_node_get_access_mode functionality so I did some digging.

That would be great.

> As far as I understand, feature nodes should only have ImposedAccessMode property and registers should only have AccessMode property. Right now feature node implements both ImposedAccessMode and AccessMode. Is there a reason for this? Maybe some camera manufacturer uses it by mistake?

You are right AccessMode should only be implemented for register nodes and register struct entry nodes.

> Current idea is to check referenced nodes using arv_gc_property_node_get_access_mode and then impose further restrictions using ImposedAccessMode if it is available.

I'm not sure it is the easiest way to handle AccessMode. There could be Converters in the path to the register nodes, so it is not easy to know which registers are affected. I guess it would be better to handle the error reported by the register nodes.

> So should I also keep AccessMode for feature node and check it in addition to ImposedAccessMode?

Outside of register nodes, only ImposedAccessMode should be handled.

EmmanuelP:
This is now in master.

PR #391, opened 2020-07-22
wip/genicam: implement 'AccessMode' and 'ImposedAccessMode' handling

sh0:
This commit implements Genicam access mode retrieval for all important features except `StructEntry` nodes.

1. In case we have a register node we get the `access_mode` property and we are done.
2. If we have non-register feature node then `arv_gc_feature_node_get_access_mode` utilizes bunch of new private functions like `arv_gc_integer_node_get_access_mode`, `arv_gc_enumeration_get_access_mode` etc. to get the referenced `pValue` node which contains wanted access mode info.

The hackiest thing about this commit is probably `arv_gc_property_node_get_access_mode`. It can take both `AccessMode` and `pValue` properties.

The approach is quite experimental, but everything seems to work. Let me know what you think.

Edit: Forgot to mention that this topic was also discussed in #332

### DIFF
--- a/src/arvgcboolean.c
+++ b/src/arvgcboolean.c
@@ -213,6 +213,22 @@ arv_gc_boolean_set_value (ArvGcBoolean *gc_boolean, gboolean v_boolean, GError *
 		g_propagate_error (error, local_error);
 }
 
+static ArvGcFeatureNode *
+arv_gc_boolean_get_pointed_node (ArvGcFeatureNode *gc_feature_node)
+{
+	ArvGcBoolean *gc_boolean = ARV_GC_BOOLEAN (gc_feature_node);
+	ArvGcNode *pvalue_node = NULL;
+
+	if (gc_boolean->value == NULL)
+		return NULL;
+
+	pvalue_node = arv_gc_property_node_get_linked_node (gc_boolean->value);
+	if (ARV_IS_GC_FEATURE_NODE (pvalue_node))
+		return ARV_GC_FEATURE_NODE (pvalue_node);
+
+	return NULL;
+}
+
 ArvGcNode *
 arv_gc_boolean_new (void)
 {
@@ -239,6 +255,7 @@ arv_gc_boolean_class_init (ArvGcBooleanClass *this_class)
 {
 	GObjectClass *object_class = G_OBJECT_CLASS (this_class);
 	ArvDomNodeClass *dom_node_class = ARV_DOM_NODE_CLASS (this_class);
+	ArvGcFeatureNodeClass *gc_feature_node_class = ARV_GC_FEATURE_NODE_CLASS (this_class);
 
 	parent_class = g_type_class_peek_parent (this_class);
 
@@ -246,4 +263,5 @@ arv_gc_boolean_class_init (ArvGcBooleanClass *this_class)
 	dom_node_class->get_node_name = arv_gc_boolean_get_node_name;
 	dom_node_class->post_new_child = arv_gc_boolean_post_new_child;
 	dom_node_class->pre_remove_child = arv_gc_boolean_pre_remove_child;
+	gc_feature_node_class->get_pointed_node = arv_gc_boolean_get_pointed_node;
 }
--- a/docs/reference/aravis/aravis-sections.txt
+++ b/docs/reference/aravis/aravis-sections.txt
@@ -632,6 +632,8 @@ arv_gc_feature_node_is_available
 arv_gc_feature_node_is_implemented
 arv_gc_feature_node_is_locked
 arv_gc_feature_node_get_visibility
+arv_gc_feature_node_get_imposed_access_mode
+arv_gc_feature_node_get_actual_access_mode
 <SUBSECTION Standard>
 ARV_GC_FEATURE_NODE
 ARV_IS_GC_FEATURE_NODE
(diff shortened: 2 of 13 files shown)


EmmanuelP:
Hi Siims,

Thanks for your work on this issue.

Here my comments about your patch:

- the arv_gc_feature_get_access_mode should be renamed to arv_gc_feature_get_actual_access_mode. Let's keep get_access_mode for really accessing "AccessMode" property.

- I don't like the hack in arv_gc_property_get_access_mode that follows the pValue. This function should only returns the "AccessMode" property value.

- A cleaner implementation would probably be use a ArvGcFeature::get_actual_access_mode virtual function.

sh0:
Ok, cool, I'll work on these issues.

BTW, should I keep `default_value` argument in all `*_get_access_mode` functions as is? In the end default value is always set as `ARV_GC_ACCESS_MODE_RW` anyways within `arv_gc_feature_node_get_access_mode`.

EmmanuelP:
If a default parameter is not necessary, please drop it.

sh0:
I have now used virtual method approach. The result is much cleaner indeed. `StructEntry` is also supported now.

When traversing `pValue` nodes I only check `AccessMode` right now. This is generally fine as feature nodes typically directly reference registers. As far as I understand, it is theoretically possible that one feature node references another feature node which also has `ImposedAccessMode` property. It is not clear to me from standards if that `ImposedAccessMode` of a `pValue` should be combined into the actual access mode or not.

src/arvgcboolean.c
```diff
@@ -213,6 +213,27 @@ arv_gc_boolean_set_value (ArvGcBoolean *gc_boolean, gboolean v_boolean, GError *
...
 }
 
+static ArvGcAccessMode
+arv_gc_boolean_get_access_mode (ArvGcFeatureNode *gc_feature_node)
+{
+	ArvGcBoolean *gc_boolean = NULL;
+	ArvGcNode *pvalue_node = NULL;
+	ArvGcFeatureNode *pvalue_feature = NULL;
+
+	g_return_val_if_fail (ARV_IS_GC_BOOLEAN (gc_feature_node), ARV_GC_ACCESS_MODE_RO);
```
EmmanuelP, line 223: Type check is useless here, gc_feature is guaranteed to be a GC_BOOLEAN.

src/arvgcboolean.c
```diff
@@ -213,6 +213,27 @@ arv_gc_boolean_set_value (ArvGcBoolean *gc_boolean, gboolean v_boolean, GError *
...
+	gc_boolean = ARV_GC_BOOLEAN (gc_feature_node);
+	if (gc_boolean->value == NULL)
+		return ARV_GC_ACCESS_MODE_RO;
+
+	pvalue_node = arv_gc_property_node_get_linked_node (gc_boolean->value);
+	g_return_val_if_fail (ARV_IS_GC_FEATURE_NODE (pvalue_node), ARV_GC_ACCESS_MODE_RO);
```
EmmanuelP, line 229: value not being a pValue is not en error.

src/arvgcboolean.c
```diff
@@ -213,6 +213,27 @@ arv_gc_boolean_set_value (ArvGcBoolean *gc_boolean, gboolean v_boolean, GError *
...
+
+	pvalue_feature = ARV_GC_FEATURE_NODE (pvalue_node);
```
EmmanuelP, line 231: Virtual should not be called directly, usually there is a function defined in the class declaring the virtual that does that.

src/arvgcboolean.c
```diff
@@ -213,6 +213,27 @@ arv_gc_boolean_set_value (ArvGcBoolean *gc_boolean, gboolean v_boolean, GError *
 		g_propagate_error (error, local_error);
 }
 
+static ArvGcAccessMode
+arv_gc_boolean_get_access_mode (ArvGcFeatureNode *gc_feature_node)
```
EmmanuelP, line 217: The content of this function is duplicated in several instances. It looks like the only difference is the access to pValue. It is a call for the creation of a get_pointed_node virtual in ArvGcFeature. May be we don't need an access_mode virtual after all.

sh0:
I now used the `get_pointed_node` approach and tried to address all the other issues as well. I hope it looks better.

EmmanuelP:
Thanks.

EmmanuelP:
Finally I have changed my mind about the pointed_node naming, and replaced it by linked_feature. That is consistant with arv_gc_property_get_linked_node, and indicates it returns specifically a ArvGcFeature.

I have pushed a patch to master that implements the renaming.

EmmanuelP:
Speaking about naming, I'm not too fond of get_actual_access_mode. I still like the fact it makes clear it does not return the value of the <AccessMode> property node, but sounds a bit weird. Do you have a better suggestion ?

sh0:
Thanks for reviewing!

As for the naming we could theoretically go back to `get_access_mode` as we only need `get_imposed_access_mode` for feature node properties. If there is a need to get `AccessMode` property for registers we could always implement a new function `arv_gc_register_node_get_access_mode` etc. Other words to use might be `get_effective_access_mode` or `get_real_access_mode`. None of the options are really great though so I also don't have a strong recommendation.

Issue #392, opened 2020-07-24
ARAVIS_FAST_HEARTBEAT does not get defined by meson configure


casperroo:
**Describe the bug**
src/arvgvdeviceprivate.h defines ARAVIS_FAST_HEARTBEAT
but
src/meson.build defines
library_config_data.set10 ('ARAVIS**_HAS**_FAST_HEARTBEAT', get_option ('fast-heartbeat'))

I am not familiar with meson but I am positive that meson.py configure build -Dfast-heartbeat=true will not produce build with teh fast heartbeat enabled

**To Reproduce**
meson.py configure build -Dfast-heartbeat=true
cd build
ninja

**Expected behavior**
build/src/arvfeatures.h should contain line like:
#define ARAVIS_FAST_HEARTBEAT whatever
but the file lacks the define

On the other hand switching for example ARAVIS_HAS_PACKET_SOCKET works as expected

**Camera description:**
n/a

**Platform description:**
 - Aravis version 0.8.0 (master: 2e4183e3492b24b9c054e4c3a8cc866dae2667a1)
 - OS: Ubuntu 18.04
 - Hardware: x86_64

**Additional context**
I tried to change the src/arvgvdeviceprivate.h to use the version with the _HAS_ but that does not help.

PR #394, opened 2020-07-24
Fix fast-heartbeat meson build option

casperroo:
the meson did set up @ARAVIS_HAS_FAST_HEARTBEAT@ variable but it was not handled in the src/arvfeatures.h.in

Fix the issue with fast-heartbit meson option desribed in issue #392


EmmanuelP:
I went for a simplified version of your patch.

Thanks.

casperroo:
Looks ckeaner bu doesn't it brake backward compatibility with 0.6?
And in an ugly way: if one relied in his/her code on #ifdef
ARAVIS_FAST_HEARTBEAT he/she won't get any warning while porting to 0.8.
Or do I get it wrong?

Chris

czw., 30 lip 2020, 11:21 użytkownik Emmanuel Pacaud <
notifications@github.com> napisał:

EmmanuelP:
I did not think about this potential issue, as I think the define is pretty useless outside of aravis compilation. But anyway, the define was not public in 0.6

casperroo:
It was not indeed. Thanks for checking.

PR #401, opened 2020-08-18
(don't merge) Add macros for checking bit depth of ArvPixelFormat

abrock:
I'd like to have a macro which tells me the actual bit depth of any ArvPixelFormat, like this:

- ARV_PIXEL_FORMAT_MONO_8 -> 8
- ARV_PIXEL_FORMAT_MONO_12 -> 12
- ARV_PIXEL_FORMAT_MONO_12_PACKED -> 12

I tried to implement and test macros which check if a ArvPixelFormat is 8bit, 10bit or 12bit but I only got the 8bit one working.

I also wrote tests for ARV_PIXEL_FORMAT_BIT_PER_PIXEL and I'm confused that it returns 12 for ARV_PIXEL_FORMAT_MONO_10_PACKED but 10 for ARV_PIXEL_FORMAT_BAYER_**_10P, is that correct?


src/arvenums.h
```diff
@@ -81,6 +81,12 @@ typedef guint32 ArvPixelFormat;
 
 #define ARV_PIXEL_FORMAT_BIT_PER_PIXEL(pixel_format) (((pixel_format) >> 16) & 0xff)
 
+#define ARV_PIXEL_FORMAT_8_BIT(pixel_format) ((pixel_format & 0x00080000) && !(pixel_format & ~0x0318003f))
```
EmmanuelP, line 84: What is wrong with ARV_PIXEL_FORMAT_BIT_PER_PIXEL ? What is the need for your new 3 macros ?

tests/arvenums.c
```diff
@@ -0,0 +1,511 @@
+#include <arv.h>
```
EmmanuelP, line 1: It is a great idea to implement more unit testing. But it should integrate with the unit testing framework. Have a lookt at evaluator.c for example.

Also, I think you went to far in the number of tests you have implemented. I'm not sure testing every defined pixel formats is useful. The tests should only exercize the bit per pixel API.

EmmanuelP (changes requested):
12 bit per pixel for ARV_PIXEL_FORMAT_MONO_10_PACKED is normal (10 effective packed to only use 24 bits for 2 pixels). I guess the 10P bayer variant is a more condensed pixel format, with no unused bits.

src/arvenums.h
abrock: ARV_PIXEL_FORMAT_BIT_PER_PIXEL gives the number of bits used during transport, e.g.:

Mono8: 8
Mono10: 16
Mono12: 16

I'd like to know the bit depth of the camera's ADC, e.g.:

Mono8: 8
Mono10: 10
Mono12: 12

I tried (and failed) to make simple macros which tell me if a ArvPixelFormat corresponds to an 8bit/10bit/12bit ADC.

EmmanuelP: Ok.

Did you read https://www.emva.org/wp-content/uploads/GenICam_PFNC_1_1_01.pdf ?

It looks like it is not possible to extract the bit depth from the pixel format 32bit identifier. I fear it has to be handled with a dedicated table, which could also include packing and location informations, based on the PFNC document.

EmmanuelP:
Unfortunately a bit depth macro seems not implementable cleanly.

I'm closing this merge request.

Issue #400, opened 2020-08-18
There is no way to relinquish control access without finalizing the whole camera


Ekleog:
**Is your feature request related to a problem? Please describe.**

A way to know whether we currently have control access has been made available with https://github.com/AravisProject/aravis/issues/382. Unfortunately, there is currently no good way to relinquish control access without finalizing the whole camera.

**Describe the solution you'd like**

Would it make sense to just make `arv_gv_device_take_control` and `arv_gv_device_leave_control` non-`static`?

**Describe alternatives you've considered**

Not doing this, but then it would mean the only way to relinquish control access on the camera is to actually finalize the whole camera, which is especially unpleasant in affine-typed languages (like Rust) where we can't have uninitialized variables hanging around.

**Additional context**

Well… as usual, thank you for all the work put on aravis! (and yes, I'll eventually find the time to come back to #396, sorry for the delay…)

EmmanuelP:
Hi,

> Would it make sense to just make arv_gv_device_take_control and arv_gv_device_leave_control non-static?

It is not as simple as that. If we want to expose this control API, we have to take care to handle correctly the heartbeat thread.

Also, it would be the opportunity to implement a lazy control access mode. ArvGvdevice would gain device control access only for write or command actions, and would release the control access after a timeout, or when the device is finalized. This default mode could exist in addition to an explicit control access.

lovesegfault:
@EmmanuelP Any idea when this can/will be landed?

Thanks!

PR #404, opened 2020-09-01
gvdevice: make it possible to manually take or leave control access

Ekleog:
Hey! I've taken a stab at implementing a minimal version of #400. What do you think?

(commit message below)

The main changes are:
- Make `arv_gv_device_take_control` and `arv_gv_device_leave_control`
  public, adding documentation
- Add an `error` parameter for consistency with the current public API
- Adjust the few callers to this new parameter
- Add a warning message if we fail to relinquish the control access
- Update `is_controller` correctly if `leave_control` failed
- Wrap writing the register in a check that the register write will
  actually have an impact

I think this is correct and does not need touching the heartbeat
thread because:
- The heartbeat thread as defined in `arv_gv_device_heartbeat_thread`
  does not do anything when `is_controller` is set to `FALSE`
- Just setting `is_controller` to `FALSE` is already what will happen
  when losing the TCP control connection
- Having one heartbeat thread per alive `GvDevice` instance sounds
  like a reasonable price to pay, even though it would probably be
  more optimal to stop the heartbeat thread on control leave and
  resume it on control take

So, as future work that could be done to further improve the
situation, I see:
- Stop the heartbeat thread when leaving control and resume it when
  taking control
- Implement the rest of #400, ie. adding a “lazy” control access mode
  whereby control access is automatically acquired/relinquished when
  doing settings only

### DIFF
--- a/src/arvgvdevice.c
+++ b/src/arvgvdevice.c
@@ -447,17 +447,30 @@ arv_gv_device_heartbeat_thread (void *data)
 
 /* ArvGvDevice implemenation */
 
-static gboolean
-arv_gv_device_take_control (ArvGvDevice *gv_device)
+/**
+ * arv_gv_device_take_control:
+ * @gv_device: a #ArvGvDevice
+ * @error: a #GError placeholder, %NULL to ignore
+ *
+ * Returns: whether the control was successfully acquired
+ *
+ * Since: 0.8.0
+ */
+
+gboolean
+arv_gv_device_take_control (ArvGvDevice *gv_device, GError **error)
 {
 	ArvGvDevicePrivate *priv = arv_gv_device_get_instance_private (gv_device);
-	gboolean success;
+	gboolean success = TRUE;
 
-	success = arv_device_write_register (ARV_DEVICE (gv_device),
-					     ARV_GVBS_CONTROL_CHANNEL_PRIVILEGE_OFFSET,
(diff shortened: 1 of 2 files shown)


Ekleog:
Hmm… I planned to heavily rely on the CI to check that my changes are correct, but it looks like the CI is broken, as meson fails to find a `mesonbuild` module.

Any idea what's going on? Locally, inside a `nix-shell -p meson pkgconfig glib libxml2 ninja`, running `meson build; cd build/; meson configure; ninja; meson test` appears to show everything working.

EmmanuelP:
> Hmm… I planned to heavily rely on the CI to check that my changes are correct, but it looks like the CI is broken, as meson fails to find a mesonbuild module.

CI should be fixed in master.

Ekleog:
Looks like CI passed this time indeed, thank you!

Ekleog:
I have just tested with an actual camera, with the following process:
1. get the camera
2. leave control
3. take control
4. configure the camera (thus confirming that the camera control is actually taken)
5. leave control
6. without finalizing the camera, get another camera, and do steps 1-4 again

Everything appeared to work correctly in practice.

lovesegfault:
@EmmanuelP Any chance this can get reviewed?

EmmanuelP:
Pushed to master.

Thanks.

Ekleog:
Thank you! :D

PR #411, opened 2020-09-23
Fix _send_cmd_and_receive_ack timeout race

casperroo:
Using MAX macro with g_get_monotonic_time() as one of parameters may
cause (and causes) a race because this is evaluated twice.
It lead to g_poll timeout being set to -1 causing infinite wait on
disconnected camera.

Very hard to reproduce. I use FAST-HEARTBEAT to make my software react for camera being plugged/unplugged faster.
After camera is lost the system polls for it once in a while in a separated thread.
I realized that arv_camera_new(camera_id) takes very long if the camera does not exists (but did exist previously). So what I am actually doing (after control-lost) is calling some read command (arv_camera_get_gain) and only once it succeeds I proceed with full re-initialization. This works fine but once in say 3 - 6 hours (!!!) with an attempt every 5 seconds the library freezes. I have compiled in in release with debug symbols and finally managed to find the spot:

`timeout_ms = MAX (0, timeout_stop_ms - g_get_monotonic_time () / 1000);`
`...`
`success = success && g_poll (&io_data->poll_in_event, 1, timeout_ms) > 0;`

gets me to the state where the timeout_ms is -1 which means "wait forever".
See the debuger (qt-creator) output.

aravis-MAX-macro-race

I scratched my head how is it possible that macro MAX(0, whatever) may return -1 but finally figured it out: double evaluation.
I believe this may have been a reason for other rare/hard-to-reporduce error. 

I have checked the remaining code and found the macro to by used correctly (without function calls as parameters) everywhere else but here:
arvgvdevice.c: 573:  inc = MAX (1, arv_device_get_integer_feature_increment (device, "GevSCPSPacketSize", NULL));

I'm not sure if the above shouldn't be fixed as well.

### DIFF
--- a/src/arvgvdevice.c
+++ b/src/arvgvdevice.c
@@ -228,7 +228,9 @@ _send_cmd_and_receive_ack (ArvGvDeviceIOData *io_data, ArvGvcpCommand command,
 			do {
 				pending_ack = FALSE;
 
-				timeout_ms = MAX (0, timeout_stop_ms - g_get_monotonic_time () / 1000);
+				timeout_ms = timeout_stop_ms - g_get_monotonic_time () / 1000;
+				if (timeout_ms < 0)
+					timeout_ms = 0;
 
 				success = TRUE;
 				success = success && g_poll (&io_data->poll_in_event, 1, timeout_ms) > 0;


EmmanuelP:
Wow, good catch !

Thanks a lot for the detailed explanation.

EmmanuelP:
> I realized that arv_camera_new(camera_id) takes very long if the camera does not exists (but did exist previously). 

That is another issue. I guess it can be fixed by removing the corresponding entry in the device list if the device creation fails.

EmmanuelP:
> I have checked the remaining code and found the macro to by used correctly (without function calls as parameters) everywhere else but here:
arvgvdevice.c: 573: inc = MAX (1, arv_device_get_integer_feature_increment (device, "GevSCPSPacketSize", NULL));

Thanks for the check. I would not mind a patch that fixes this occurence. :)

EmmanuelP:
>  >   I have checked the remaining code and found the macro to by used correctly (without function calls as parameters) everywhere else but here:
    arvgvdevice.c: 573: inc = MAX (1, arv_device_get_integer_feature_increment (device, "GevSCPSPacketSize", NULL));

> Thanks for the check. I would not mind a patch that fixes this occurence. :)

Don't bother, patch incoming.

casperroo:
> Wow, good catch !
> 
> Thanks a lot for the detailed explanation.

Thanks for taking a time to read this. And thanks for quick merge.

casperroo:
> > I realized that arv_camera_new(camera_id) takes very long if the camera does not exists (but did exist previously).
> 
> That is another issue. I guess it can be fixed by removing the corresponding entry in the device list if the device creation fails.

This is not what I need. I do need to keep the entry in a list. I have a camera manager which holds individual cameras (in seperated threads). When I wand to remove/refresh the camera list i do it from there. 
But in my scenario it is quite common for the camera to be disconnected for long period of time (say over a night) and I am expecting the camera to reappear. Our customers usually power off machine for a night without powering the computers that the cameras are attached. When i happens each camera thread holds the device_id string and polls for the camera to reapear. Initially that was done in the non-fast-heartbeat version by calling arv_camera_new. If the camera does not exists it takes 12 seconds (!!!). With fast-heartbeat 6 or so (still much). So that is why i use get_gain as kind of ping. 
Maybe it would be good to have some "ping" command that does some minimal operation by say checking if the camera's port is open.

EmmanuelP:
>  If the camera does not exists it takes 12 seconds (!!!). With fast-heartbeat 6 or so (still much). 

That is an issue, we should have a much shorter timeout.

PR #414, opened 2020-10-07
16bit fake image

hinxx:
I've playing around with the fake camera and added support for generating 16 bit images. Might be worthy of adding to the upstream.


EmmanuelP:
Hi Hinko,

Thanks for the patch. Unfortunately, it doesn't pass the test suite, as indicated by the travis CI failure. Could you fix this issue ?

EmmanuelP:
Thanks.

PR #423, opened 2020-10-19
Inline fallback fake camera XML

eudoxos:
This is an attempt to improve relocability of the binaries by inlining the fallback fake camera XML into the source itself.

This concerns `arv-viewer` in particular, which does not have the option to pass XML path directly and does not honor `XDG_DATA_DIRS` and similar, making cross-compilation difficult. `ARAVIS_DATA_DIR` is hard-coded at build-time.


EmmanuelP (changes requested):
I'm open to the goal of this merge request, but please use  https://developer.gnome.org/gio/stable/GResource.html . I don't how if it plays well with cross compilation though.

At least the solution should be integrated  to the build system (no manual intervention, no intermediate file stored in git).

eudoxos:
This would pull in the gnome module of meson [gnome.compile_resources](https://mesonbuild.com/Gnome-module.html#gnomecompile_resources), are you okay with that? That would also allow introspection to use [gnome.generate_gir](https://mesonbuild.com/Gnome-module.html#gnomegenerate_gir), fixing cross-compilation issue for that module (hopefully).

EmmanuelP:
> This would pull in the gnome module of meson gnome.compile_resources, are you okay with that? That would also allow introspection to use gnome.generate_gir, fixing cross-compilation issue for that module (hopefully).

This is already the case. meson gnome module is imported unconditionally, and compile_resources is used for the viewer.

generate_gir is used if introspection is enabled.

eudoxos:
Continue in #426.

PR #426, opened 2020-10-21
Embed fake camera genicam XML as GResource (instead of external file)

eudoxos:
As discussed in #423 . It was easier than expected, cross-compilation not tested yet.


EmmanuelP:
Tests don't pass:

eudoxos-gres [$] ❯ ninja test
[0/1] Running all tests.
1/8 aravis:main / evaluator      OK             0.02s
2/8 aravis:main / buffer         OK             0.03s
3/8 aravis:main / misc           OK             0.02s
4/8 aravis:main / genicam        OK             0.05s
5/8 aravis:main / fake           FAIL           1.13s (killed by signal 6 SIGABRT)
6/8 aravis:network / fakegv      FAIL           1.68s (killed by signal 6 SIGABRT)
7/8 aravis:python / fake.py      FAIL           0.18s (exit status 1)
8/8 aravis:python / exception.py FAIL           2.63s (exit status 1)

Ok:                 4   
Expected Fail:      0   
Fail:               4   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0

eudoxos:
I obviously did not grep in `tests`; and a reminder for me to use the tests what they're for. Sorry. It is fixed now.

It should be possible to use `resource:///org/gnome/aravis/arv-fake-camera.xml` as filename in GLib functions, that it did not work for some reason it did not work. So for now use the XML from the source tree.

As a suggestion, when XML is not found, should not we produce a lounder warning? arv_warning_device outputs only when `ARV_DEBUG` is set appropriately.

EmmanuelP:
> As a suggestion, when XML is not found, should not we produce a lounder warning? arv_warning_device outputs only when ARV_DEBUG is set appropriately.

Yes you are right. g_critical seems more appropriate.

eudoxos:
I added ;g_critical; and also created a few spurious commits with `pull --rebase`. I hope you can flatten the PR when merging.

EmmanuelP:
Commited to master. Thanks.

Issue #429, opened 2020-10-28
Device with ?SchemaVersion in the XML URL isn't parsed correctly


JoseJX:
I tried to use a Mikrotron camera with Aravis. Running arv-tool produced the following output:
`Failed to open device 'Mikrotron GmbH-MC2064-0000000000000009': Invalid Genicam data
`
After some inspection, I found the issue was the XML URL and the SchemaVersion at the end:
`Local:Mikrotron_GmbH_MC206xS11_Rev0_00_007.zip;8001000;395C?SchemaVersion=1.1.0
`

After editing the string manually in gdb and inserting a NULL where the ? was, I was able to connect to the camera and start a stream.

Camera details are:
 - Manufacturer: Mikrotron GmbH
 - Model: EoSens
 - Interface: GigE Vision

**Platform description:**
 - Aravis version: Latest Git (9ac7e7e055386b9525d77bddf5799d330711303f)
 - OS: Gentoo
 - Hardware x86_64

JoseJX:
Just wanted to say that the merged fix lets the Mikrotron camera above work great with Aravis. Thanks!

PR #430, opened 2020-10-28
gvdevice: Add support for URLs with SchemaVersion.

JoseJX:
Fixes bug #429 by simplifying the parsing process. I also changed the tests to reflect the changed output from the URL parser. I'm not sure this is the correct fix, but it works for me with the Mikrotron camera referenced and an Allied Vision camera.


EmmanuelP:
Hi Joe,

Thanks for the report and the pull request. In the end I have implemented a fix that isolate the parsing part here: #433

Issue #416, opened 2020-10-09
Cleaning up to make mingw64 compilation straightfoward


eudoxos:
Far from being a fan of the Windows platform, I am under pressure to compile a code using Aravis for it. Using [crossroad](https://pypi.org/project/crossroad/) makes cross-compilation with gcc/mingw64 and meson (and installation of all dependencies via msys2 packages) feasible.

The compilation of aravis currently fails in several places, mostly because of platform-specific headers.  I was able to work around most of them, but would like to have some guidance and a go-ahead from the project so that those few changes can be merged later:

* Some header inclusion should be guarded (`#ifndef __MINGW32__` or similar; this concerns e.g. `sys/resource.h` and `sys/syscall.h`), similar to `#ifndef __APPLE__` which already appears in a few places.
* Some headers would conditionally use alternatives (`sys/socket.h` → `winsock.h`, for `setsockopt`).
* Some non-essential components using non-portable headers (`arvgvfakecamera.c`) would be excluded from the build.
* Some tests (`test/realtimetest.c`) would trivially fail.
* Features like documentation, viewer, introspection, usb will not be cross-compilable (that is okay, for the start or forever).

The first step would be to make the code compile (for all platforms, incl. mingw), later test how functional it is on a native platform.

Could you suggest what preprocessor macros to use to guard platforms? It seems that the mingw compiler unfortunately predefines `_POSIX_VERSION` (which [should normally be used to detect POSIX](https://stackoverflow.com/a/11351171/761090)) despite not providing POSIX `sys/...` headers. Perhaps combining `__APPLE__` and `__MINGW32__` (which is also defined by mingw64) would be sufficient, for the start.

NB: this is **not** about compiling Aravis using MSVC.

EmmanuelP:
Hi,

Thanks for your work on this.

I'm perfectly fine with a patch set that makes aravis compile on Windows, even with a reduced feature set. It would be nice if you could make also fakecamera compilable though.

There is a glib macro that detects Windows, G_OS_WIN32. I don't know if that help.

It would also be great if you could manage to configure a CI pipeline for the Windows build.

eudoxos:
The last missing bit seems to be `getifaddrs` which is non-portable. Some of the parts could be replaced by more portable `getaddrinfo`, except of the bit where netmask is queried (I don't know how to do that). What do you prefer? Use `getifaddrs` for unices and native windows API in windows, or do as much as possible via unified code with `getaddrinfo`?

[...]

PR #441, opened 2020-11-03
mingw: only define inet_ntop conditionally, depending on _WIN32_WINNT

eudoxos:
This should solve `inet_ntop` conflict described in https://github.com/AravisProject/aravis/issues/416#issuecomment-717220610 by @NerdToMars


EmmanuelP:
Thanks.

PR #448, opened 2020-11-26
mingw: replace M_PI and M_E with G_PI and G_E

sh0:
Constants M_PI and M_E are not defined in standard C. This will result compilation errors under MinGW (x86_64-w64-mingw32-gcc, version 10.2.0). Fortunately glib defines G_PI and G_E so use those instead.


kalvdans (approved):
Looks good to me!

EmmanuelP:
Thanks.

PR #455, opened 2020-12-02
camera: also handle GainRaw as float feature

EmmanuelP:


JiangXL:
It does work. 
Thanks.

EmmanuelP:
Thanks for the test.

PR #468, opened 2021-02-09
Action API support

EmmanuelP:


EmmanuelP (commented):
I'm not fond of the current API. It is too low level.

May be a better approach would be to configure the action with just giving a set of devices. We may be able to automatically define the needed keys, and have a `*_full` version with user given key values. I guess the broadcast address can be also found automatically, or set explicetely in the `*_full` API version.

PR #471, opened 2021-02-16
Fix arv_camera_set_trigger for Basler

casperroo:
AcquisitionFrameRateEnable is boolean node rather than integer

See: https://docs.baslerweb.com/acquisition-frame-rate
In fact everywhere else in the code the AcquisitionFrameRateEnable is used by means of the integer node.

To be honest I do not understand why the arv_camera_set_trigger disables AcquisitionFrameRateEnable only for Basler cameras but nevertheless without this fix the function fails anyway.


EmmanuelP:
Hi,

Thanks for the merge request. Your change is now in master.

Cheers.

PR #472, opened 2021-02-22
retrieve timeouts, missing frames/packets and resend limits

spatel-cog:
I'd like more detailed stats to be available via the python interface. 
Please let me know if these changes are acceptable or whether the current get_statistics API should be modified to include additional stats. I didn't want to break older code hence the introduction of a new API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ addons:
     update: true
 
 before_install:
-  - sudo pip3 install --upgrade setuptools pip
+  - sudo pip3 install --upgrade setuptools pip==20.3.4
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then sudo pip3 install ninja ; fi
   - sudo pip3 install 'meson==0.47'
 
--- a/src/arvgvstream.c
+++ b/src/arvgvstream.c
@@ -1261,11 +1261,10 @@ _get_detailed_statistics(ArvStream *stream,
 	thread_data = priv->thread_data;
 
 	*n_timeouts = thread_data->n_timeouts;
-        *n_missing_frames = thread_data->n_missing_frames;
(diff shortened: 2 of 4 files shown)


EmmanuelP:
Hi Sandip,

Thanks for the merge request.

As you have noted, the current implementation is not flexible and require an API change each time we want to add a new data. Also, some data only make sense for a given interface.

So I would propose the following API instead of adding another non extensible one:

guint arv_stream_get_n_infos (ArvStream *stream)

double arv_stream_get_float_info (ArvStream *stream, guint id)
gint64 arv_stream_get_integer_info (ArvStream *stream, guint id)

const char *arv_stream_get_info_name (ArvStream *stream, guint id)
double arv_stream_get_float_info_by_name (ArvStream *stream, const char *name)
gint64 arv_stream_get_integer_info_by_name (ArvStream *stream, const char *name)

May be a bit over engineered, but extensible, and give us the possibility to add framerate and bandwidth informations.

EmmanuelP:
By the way, I have commited the CI fix. It is in master now.

Thanks.

spatel-cog:
Hi Emanuel, 
Thanks for reviewing my change request. 
I like your idea of having APIs to read ArvStream specific values. This will go a long way to exposing the kind of information that higher level APIs would like to see. 
Let me know how I can help.

Regards,
Sandip Patel,
Cogniac Corporation,
160 West Santa Clara St., Suite 1575
San Jose CA 95113.

EmmanuelP:
> Let me know how I can help.

Well, if you have the time and willingness, you may try to implement the proposed API.

EmmanuelP:
I have implemented an extensible stream info API here: https://github.com/AravisProject/aravis/pull/523

It obsoletes this pull request. I'm therefore closing it.

Thanks.

PR #478, opened 2021-03-03
Fix glib deprecation warning

EmmanuelP:
To be merged when glib 2.68 is released.

### DIFF
--- a/src/arvmisc.h
+++ b/src/arvmisc.h
@@ -71,6 +71,12 @@ ArvPixelFormat 	arv_pixel_format_from_gst_0_10_caps 		(const char *name, int bpp
 
 const char *	arv_vendor_alias_lookup		(const char *vendor);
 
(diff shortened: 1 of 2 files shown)


kalvdans:
Just silence deprecation warnings if we need to be backwards compatible with old glib version.

src/arvmisc.h
```diff
@@ -71,6 +71,12 @@ ArvPixelFormat 	arv_pixel_format_from_gst_0_10_caps 		(const char *name, int bpp
 
 const char *	arv_vendor_alias_lookup		(const char *vendor);
 
+#if GLIB_CHECK_VERSION(2,68,0)
+#define arv_memdup(p,s) g_memdup2(p,s)
+#else
+#define arv_memdup(p,s) g_memdup(p,(size_t) s)
```
kalvdans, line 77: `g_memdup(p,(size_t) (s))`

PR #520, opened 2021-06-22
Fixing bounds for clamping of exposure mode

mtnwrw:
The clamping for the strings on the exposure mode had a wrong range, it always clamped to 0 (==Off). 
Single-line fix.


EmmanuelP:
Thanks for the pull request. I have pushed a tweaked version of your patch.

PR #521, opened 2021-06-29
Print number of detected devices in arv-tool

abrock:
I sometimes have 18 cameras connected and sometimes one is not responding. In this case it's very useful to immediately get the number of responding cameras.

### DIFF
--- a/src/arvtool.c
+++ b/src/arvtool.c
@@ -485,6 +485,7 @@ main (int argc, char **argv)
 		n_devices = arv_get_n_devices ();
 
 		if (n_devices > 0) {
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Hi,

Thanks for the pull request. I'm not a huge fan of this additional line, especially now the complete list is displayed instantly when no action is required.

`arv-tool-0.8 | wc -l` should give you the number of found devices.

abrock:
I like to see an overview of the cameras, so using | wc -l would require me to run arv-tool twice.

EmmanuelP:
`arv-tool-0.8 | cat -n` ?

abrock:
That's very cool, I didn't know that flag, thank you.

PR #534, opened 2021-07-22
Action for automated mingw build

eudoxos:
Preview only; aravis:main / fakegv test fails for some unknown reason.

### DIFF
--- a/.github/workflows/aravis-mingw.yml
+++ b/.github/workflows/aravis-mingw.yml
@@ -55,4 +55,4 @@ jobs:
         ninja -C ./build --verbose install
     - name: ninja test
       run: |
-        ninja -C ./build --verbose test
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Great ! Thanks a lot.

The fakegv test tries to instantiate a fake GigEVision camera, so I guess it fails for the same reason `arv-fake-gv-camera` fails.

EmmanuelP:
fakegv can be ignored using `meson test --no-suite="network"` as the test command.

eudoxos:
Okay, it runs now successfully. Pull this branch or just copy the workflow definition, up to you. The workflow itself documents how to build under windows (with msys2): install dependencies, run `meson`, run `ninja install`, finito.

PR #536, opened 2021-07-23
fakecamera: allow to specify listening interface by address

### LINKED ISSUE #535: fake camera: specify IP address to bind rather than network interface name
**Is your feature request related to a problem? Please describe.**
The current fake camera has the `interface-name` property, the default `lo` is not portable. Camera setup cycles through interfaces only to obtain the right IP address to bind.

**Describe the solution you'd like**
Instead of specifying `interface-name`, IP address (or hostname) should be specified. The default could be `127.0.0.1`. The loop over interfaces could be also removed. Most UNIX servers specify IP adress they bind, so it would be culturally consistent.

**Describe alternatives you've considered**
Finding loopback interface name under windows (and use that as the default) -- until I found that loopback (mostly) does not even exist as network interface under Windows.

Another alternative is to have the option to specify either the interface or the IP address. This would probably be superfluous and add unnecessary complexity for little use.

**Additional context**
Win32 port is the incentive to bring this up.

EmmanuelP:
And default to '127.0.0.1' instead of 'lo'.

Fixes #535

### DIFF
--- a/.github/workflows/aravis-ci.yml
+++ b/.github/workflows/aravis-ci.yml
@@ -53,7 +53,7 @@ jobs:
       env:
         CC: gcc
     - name: Tests
-      run: meson test -C build/ -v --no-suite="network"
+      run: meson test -C build/ -v
     - uses: actions/upload-artifact@v2
       if: failure()
       with:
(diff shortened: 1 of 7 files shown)


eudoxos:
This will still fail in Windows as there is no loopback interface reported at all (even though networking via 127.0.0.x works).

PR #538, opened 2021-07-24
Add arv_network_get_interface_by_name, use it in fake camera

eudoxos:
The intent is to make it eventually possible to use also IP address (rather than interface name) in fake camera, and this will make it easier. Perhaps such function would be also used elsewhere where interface is looked up by name, but I did not check much.

### DIFF
--- a/.github/workflows/aravis-mingw.yml
+++ b/.github/workflows/aravis-mingw.yml
@@ -56,4 +56,4 @@ jobs:
         ninja -C ./build --verbose install
     - name: meson test
       run: |
-        meson test -C ./build --no-suite="network"
+        meson test -C ./build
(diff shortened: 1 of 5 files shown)


eudoxos:
With the last commit, I can run fake camera in (virtualboxed) windows and discover it from arv-viewer on the Linux host (windows is somehow not happy about the global & subnet broadcasts; I made those non-fatal).

EmmanuelP:
Hi,

It is a bit unfortunate we started the same work at the same time. I took the lazy route and pushed my branch, your pull request being a bit more complex than what I've done.

I hope you don't mind too much. :(

Thanks anyway.

eudoxos:
No problem, but please take the new bits in arvnetwork.c - `arv_network_get_interface_by_address`, `arv_network_get_interface_by_name`, `arv_network_get_fake_ipv4_loopback` and `arv_network_interface_is_loopback`. I think it will still be useful.

eudoxos:
Or just leave it, I will put them into a separate PR later. Sorry for the overlap.

PR #550, opened 2021-08-08
msvc: MSVC compilation experiment

sh0:
I have created an experiment to try to compile Aravis on MSVC 2019. I have gotten to the point that the main library and tools compile and generally work. Viewer and gst plugins have not yet been tried out. Unfortunately some sacrifices have to be made to get MSVC support such as dropping all `g_auto*` usage.

Details about changes are in commit messages so please see that for more information.

I have also created a recipe for building the library in Conan package manager [here](https://github.com/sh0/conan-center-index/tree/aravis-package/recipes/aravis/all). This is particularly useful when trying to compile Aravis on Windows machine.

I think it is worth considering adding MSVC support even though some changes are pretty nasty. Let me know what you think.


sh0:
Currently the tests in Aravis contain lot of `g_auto` references that I did not want to spend time working on so those tests were not done on Windows.

CI compile error on MacOS is due to a small bug in my `arv_rtkit_*` patch.

Windows `arv-test` log:
```
Testing device 'Fake' from 'Aravis'
                   GenicamLoad SUCCESS
                 GenicamSchema IGNORED
             SensorSizeReadout SUCCESS
               SensorSizeCheck SUCCESS 2048x2048
           MultipleAcquisition FAILURE
  MultipleAcquisitionFrameRate FAILURE
             CameraAcquisition FAILURE
```

Windows `arv-camera-test` log (done on a slow VM):
```
Looking for the first available camera
vendor name            = Aravis
model name             = Fake
device serial number   = GV01
image width            = 512
image height           = 512
horizontal binning     = 1
vertical binning       = 1
exposure               = 10000 ┬Ąs
gain                   = 0 dB
payload                = 262144 bytes
gv n_stream channels   = 1
gv current channel     = 0
gv packet delay        = 0 ns
gv packet size         = 1400 bytes
  0 frame/s  -       0 MiB/s - 24 errors
  0 frame/s  -       0 MiB/s - 25 errors
  0 frame/s  -       0 MiB/s - 25 errors
  0 frame/s  -       0 MiB/s - 25 errors
n_completed_buffers    = 0
n_failures             = 99
n_underruns            = 0
n_timeouts             = 0
n_aborteds             = 0
n_missing_frames       = 0
n_size_mismatch_errors = 0
n_received_packets     = 15735
n_missing_packets      = 9017
n_error_packets        = 0
n_ignored_packets      = 0
n_resend_requests      = 0
n_resent_packets       = 0
n_resend_ratio_reached = 0
n_resend_disabled      = 0
n_duplicated_packets   = 0
n_transferred_bytes    = 21467256
n_ignored_bytes        = 0
```

So there are issues, but it is meaningful to work on them only if we are going ahead with this port.

EmmanuelP:
Hi Siim,

Thanks a lot for your work on this MSVC port.

There is 3 different aspects in this port I'm not fond of:

- the addition of ARV_API. I tend to prefer clean header files. But I guess it is a good mean to avoid leaking private API, like the function declared in the private headers.
-  the addition of an initialization function. It looks it is possible to avoid it, like what was done in glib: [https://bugzilla.gnome.org/show_bug.cgi?id=752837](https://bugzilla.gnome.org/show_bug.cgi?id=752837)
- the removal of the `g_auto_*` calls. This one is a bit more problematic, as I tend to use them more and more. But I think I can live without it.

No real show stoppers so far.

I don't know much about development on Windows, so the question is, before going further: what is the added value to this port with respect to the mingw port ? Do we expect better performance ? Do we just gain a better Visual Studio integration ?

sh0:
Hi! Thanks for taking a look and I agree with your points. Some ideas:

* The `ARV_API` can theoretically be avoided if we craft a list of symbols to export and provide them to MSVC. However we then need to keep the list up-to-date and the list format is MSVC-specific. Most projects seem to have opted to go with `__declspec(dllexport)` for maintainability reasons. I do agree that this macro is pretty ugly.
* I have to investigate that glib trick for initialization. If we can use it then great.
* Perhaps in some later MSVC the `g_auto_*` functionality becomes available as people have been [requesting it](https://developercommunity.visualstudio.com/t/add-support-for-gcc-like-cleanup-attribute-for-pla/1187001), but who knows when it will materialize. Real sad situation.

The MSVC vs MinGW question answer probably depends a lot who you ask and what their particular use case is. For me the reasons are mostly toolchain related:

* Right now, for a project that I'm doing, Aravis is the only dependency that I cannot build within MSVC. So setting up MinGW just for that is sort of painful, but doable. Also setting up MinGW on Windows has in my experience always been a mess compared to cross-compiling. Having multiple build hosts is again kinda troublesome.
* For some libraries that use Windows SDK or use C++ in non-open-source binaries there is no other way than to use MSVC. That means I really cannot go with pure MinGW environment either.
* The dependencies (glib, zlib, libxml2, libusb) are also used by other other libraries in my project. Since they all have pure C API I can generate them with either compiler and they should work seamlessly. In reality there are always some issues. Sometimes library naming is messed up, sometimes some symbol cannot be found etc. From experience it is easier if I stick with one toolchain.
* MSVC debug information PDB cannot be created in MinGW. Having debug files for all dependencies can be pretty useful, but again this requires everything to be compiled with same compiler.
* MinGW needs its own runtime library (libgcc). Fortunately that can be statically linked in though. Sometimes different standard libraries also cause issues when libraries do bad things with memory management etc across library boundaries. Fortunately it does not look like there are any issues like that currently in Aravis.

I don't think there is a big performance difference though I have no benchmark information. In the end it is just convenience and build system simplicity for me.

EmmanuelP:
Ok.

So the requirements for this pull request to be merged are:

- [] No library initialization function
- [] Remove all uses of `g_atuto*` functions
- [] All tests should pass
- [] Put MSVC build under CI control as a github action

The last item is really important as I don't and probably will never use MSVC. Without CI, this port will bitrot unavoidably.

sh0:
Agreed. It will probably take some time for me to get everything done.

sh0:
This might be a good time to take another look at my patchset. All requirements should be filled, though with minimal effort.

The CI workflow only builds `x86_64` binary and does not have introspection or GTK/gstreamer-based viewer application support. The reason is that building those dependencies does currently not work on Conan package manager on Windows.

One option is to cherry-pick some commits to master to reduce number of commits that I have to maintain.

EmmanuelP:
Thanks @sh0 

The patch series breaks the documentation building. It looks like it is caused by the ARV_API prefix, and can be fixed by moving the prefix just above the function prototype.

Also, I'm seeing a crash at the end of arv-test-0.8:

```
Found 1 device
Testing 'Point Grey Research:Blackfly S BFS-U3-13Y3C'
Genicam:Load                        SUCCESS 
Genicam:Schema                      SUCCESS Version check disabled
Properties:SensorSizeReadout        SUCCESS 
Properties:SensorSizeCheck          SUCCESS 1280x1024
Properties:GainReadout              SUCCESS 
Properties:ExposureTimeReadout      SUCCESS 
MultipleAcquisition:BufferCheck     SUCCESS 10/10
MultipleAcquisition:FrameRate       SUCCESS 10.00 Hz
SingleAcquisition:BufferCheck       SUCCESS 
Chunks:BufferCheck                  SUCCESS 
SoftwareTrigger:*                   IGNORED 
 Disable SoftwareTrigger. We get 'USB transfer error: LIBUSB_ERROR_TIMEOUT' on software trigger emission
Erreur de segmentation (core dumped)
```

sh0:
Thanks for the review.

I found `--ignore-decorator` argument for `gtkdoc-scan` that can be used to resolve the documentation issue. I initially placed the `ARV_API` on separate lines, but this tended to make headers very long so I wasn't sure it is good in terms of style.

As for the segfault I should probably go over the `g_auto` removal commit as that is most likely to be the source of problems. Perhaps you can do a stack dump - that would help. Unfortunately I don't have any USB3 cameras right now, but I'll try to test with what I have.

EmmanuelP:
> I found `--ignore-decorator` argument for `gtkdoc-scan` that can be used to resolve the documentation issue. I initially placed the `ARV_API` on separate lines, but this tended to make headers very long so I wasn't sure it is good in terms of style.

If `--ignore-decorator` allows to restore the documentation build and keep ARV_API on the same line as the function prototypes, that's great.

> As for the segfault I should probably go over the `g_auto` removal commit as that is most likely to be the source of problems. Perhaps you can do a stack dump - that would help. Unfortunately I don't have any USB3 cameras right now, but I'll try to test with what I have.

The segfault happen with any cameras, even the fake one.

sh0:
As suspected the segfault was a mistake in my `g_auto` removal commit. That is probably the most dangerous commit in this set, but I went over it once more and I hope all of it is correct now.

Documentation also works.

PR #551, opened 2021-08-10
tests: Compile 'realtimetest.c' only on Linux.

sh0:
Current `realtimetest.c` uses private functions from `arvrealtimeprivate.h` that are Linux RealtimeKit specific. Therefore disable this test on other platforms.

Potentially another more high-level test could be written that uses public `arv_make_thread_realtime` and `arv_make_thread_high_priority` functions. That could be enabled on all platforms.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -71,8 +71,8 @@ if get_option('tests')
 		['cpp-test',			'cpp.cc']
 	]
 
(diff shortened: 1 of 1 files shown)


EmmanuelP:
This code was only meant as a test bed when I was playing with the different ways of increasing the stream thread priority on linux.

Thanks for the patch.

PR #560, opened 2021-08-27
Fix setting of trigger for FrameStart

hgaiser:
Since `0.8.16`, setting the `TriggerSource` to software results in the following error on our Mako camera:

```
Value not found in <Enumeration> 'TriggerSource'
```

This issue seems to be introduced in [this](https://github.com/AravisProject/aravis/commit/f6d5f6077bcf56d837e4a6bf0eb58cbe64ed3ffa#diff-d7be5be1f918127f4161b7f5a654290347f3334e434d14a61eb04e013b85dd99) commit.

It occurs when both `AcquisitionStart` and `FrameStart` are possible trigger selectors. When both are available, the trigger selector is first set to `FrameStart` and the trigger is enabled. Then, the selector is set to `AcquisitionStart` and the trigger is disabled. Finally, the `TriggerSelector` is set to the chosen source, but the selector is still set to `AcquisitionStart`.

This PR sets the `TriggerSelector` back to `FrameStart` after having disabled the `AcquisitionStart` trigger.

### DIFF
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -1259,28 +1259,31 @@ arv_camera_set_trigger (ArvCamera *camera, const char *source, GError **error)
 
 	if (local_error == NULL &&
            arv_camera_is_enumeration_entry_available (camera, "TriggerSelector", "FrameStart", &local_error)) {
-                arv_camera_set_string (camera, "TriggerSelector", "FrameStart", &local_error);
-                if (local_error == NULL)
-                        arv_camera_set_string (camera, "TriggerMode", "On", &local_error);
-        } else {
-                has_frame_start = FALSE;
-        }
+		arv_camera_set_string (camera, "TriggerSelector", "FrameStart", &local_error);
+		if (local_error == NULL)
+			arv_camera_set_string (camera, "TriggerMode", "On", &local_error);
+	} else {
+		has_frame_start = FALSE;
+	}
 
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Thanks Hans.

I will try to review your changes until the end of this week.

EmmanuelP:
I have a chance to test your patch today.

The trigger configuration is a bit fragile, and while your patch fixes the issue, it breaks the trigger setting on 3 cameras I have for tests (2 USB3Vision Basler Ace and 1 Lucid GigEVision).

I have to spend some time to figure out how to make this code works for everyone.

de-vri-es:
I think another problem now it that the code only checks if a trigger selector is available, but not if the trigger supports the desired source. Ideally the code would also check which trigger selectors exist *and* support the desired source, then disable the other trigger and configure the wanted one.

EmmanuelP:
Thanks @de-vri-es for the suggestion.

@hgaiser, please try this branch: https://github.com/AravisProject/aravis/tree/wip/set-trigger

EmmanuelP:
I have pushed my branch to master. Thanks for the merge request, it helped to improve trigger setting in Aravis. arv_camera_set_trigger should be more robust now.

hgaiser:
Apologies for not responding earlier, I was out of office. Happy to see it is resolved, thank you @EmmanuelP !

PR #565, opened 2021-09-03
Utilize U3V device GUID instead of serial number to identify the device

iitaku:
I've encountered devices which has the same serial number. 
In that case, it is impossible to determine the device by aravis. 
I have found U3V device GUID is required to be globally unique. 
This PR makes change to utilize U3V device GUID to identify the device.


iitaku:
I want to describe the background a bit more about this PR.

I'm working with the camera unit which has two image sensors and one control unit that runs as a stereo vision camera unit. It provides two USB interfaces for each image sensor then this camera unit is recognized as two devices by Aravis.

It provides two different U3V Device GUID to identify two image sensors but it has one serial number. 
The serial number is manufacturer specific field so it is better to use U3V Device GUID to identify the device.

This PR implements the feature to identify the device by U3V Device GUID.
Can you please give some feedback for this? @EmmanuelP

EmmanuelP:
Hi @iitaku 

Thanks for the explanation. I understand better why there is 2 devices with the same serial number.

Unfortunately, we can not just simply change the naming of the devices, as it will break the setup of people relying on the current naming scheme used until now.

I had a similar issue with GigEVision devices with 2 network interfaces, seen as 2 different GigEVision devices, with the same serial number, but with a different model number. What I did in ArvGvInterface was to register the same device with 2 names, one build using manufacturer-serial, like it used to be, and a new one with manufacturer-model-serial. Which means the code still worked for people using the simpler naming scheme.

The USB3Vision code should do the same, and register a device with both manufacturer-serial and manufacturer-guid-serial.

 As a bonus, we could also register the device with a USB bus path string (not sure if there is standard way to express a BUS path), which would allow to reach whatever device is connected to a given USB port.

Fixstars-iizuka:
@EmmanuelP Thanks for your comment.

I will take a look GigEVision implementation and follow the same manner not to break the existing device naming rules.

iitaku:
@EmmanuelP Thanks for your comment.

I will take a look GigEVision implementation and follow the same manner not to break the existing device naming rules.

iitaku:
@EmmanuelP I have fixed ArvUvInterface to accept `manufacturer-guid-serial` along with the existing device naming scheme. Can you please review it?

EmmanuelP:
Thanks @iitaku for the updated patch.

It looks fine now, except may be for the arv_uv_device API break. I will do a complete review soon.

EmmanuelP:
Hi @iitaku 

I have created a new pull request with your work and some tweaks of mine. In this pull request, the API break is fixed, and I have added a new ArvUvDevice constructor that takes only a USB3Vision GUID, which, if I have understood correctly, should be enough to unambiguously identify a device.

Could you please test it ?

There is probably one last thing to fix, which is the viewer. Currently, with your device, I'm guessing you have 2 lines but it always select the same one.

iitaku:
@EmmanuelP I have tested wip/usb-guid branch and confirmed it works perfectly. Additionally, the viewer works as intended!

EmmanuelP:
Ok, great. Thanks for the test.

PR #566, opened 2021-09-03
Add TriggerSoftware support to fake camera

romainreignier:
Preliminary work to introduce TriggerSoftware support in the fake camera

Any comment is welcomed.

Remaining issues:

* [ ] implement `TriggerSelector`. `arv_camera_set_trigger` uses `TriggerSelector` to first enable `TriggerMode` on `FrameStart` and then disable it on `AcquisitionStart`. But the last `TriggerMode` value is `Off` my condition in `arv_fake_camera_is_in_software_trigger_mode` is `false`. An easy solution is to remove the `AcquisitionStart` feature from the XML but there is a test against it in `fake.c`

* [ ] Do we need to check `ARV_FAKE_CAMERA_REGISTER_TRIGGER_SOFTWARE` against the value `1` or `>0`? And to acknowledge, setting it to `0` or decrement it until `0`? Because `arv_gc_command_execute` seems to increment the value.

### DIFF
--- a/src/arv-fake-camera.xml
+++ b/src/arv-fake-camera.xml
@@ -258,6 +258,11 @@
 		<pFeature>AcquisitionMode</pFeature>
 		<pFeature>AcquisitionStart</pFeature>
 		<pFeature>AcquisitionStop</pFeature>
+		<pFeature>TriggerSelector</pFeature>
+		<pFeature>TriggerMode</pFeature>
+		<pFeature>TriggerSoftware</pFeature>
+		<pFeature>TriggerSource</pFeature>
+		<pFeature>TriggerActivation</pFeature>
 		<pFeature>ExposureTimeAbs</pFeature>
 	</Category>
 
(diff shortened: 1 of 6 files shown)


romainreignier:
@EmmanuelP I am not sure about how to handle the selector in the fake camera. Do you have an idea on this?
In the `arv_fake_camera_write_memory` function, we would need something to write the value in different registers in case of a selector.

EmmanuelP:
> @EmmanuelP I am not sure about how to handle the selector in the fake camera. Do you have an idea on this?
> In the `arv_fake_camera_write_memory` function, we would need something to write the value in different registers in case of a selector.

No need to modify `arv_fake_camera_write_memory`. Have a look at the Genicam specification. For example, the address of a register node can be a pointer to a swissknife, that coud take the value set by TriggerSelector in order to compute an indexed address. But there is probably other means to achieve the same result. Also TriggerSelector does not have to be a real register, the value can be stored in a <Integer> node.

https://www.emva.org/wp-content/uploads/GenICam_Standard_v2_1_1.pdf

romainreignier:
Thank you, I am not familiar with this swissknife concept, I will have a look at the standard.

EmmanuelP:
Pushed to main.

Thanks.

Issue #586, opened 2021-09-24
arv_camera_*_brightness() methods added to ArvCamera


hackel6:
**Is your feature request related to a problem? Please describe.**
Needing to use more generic arv_camera_get_<type>() methods to access register values associated with brightness:
`Brightness, BrightnessAbs, BlackLevel, BlackLevelRaw`

**Describe the solution you'd like**
Implementation of arv_camera_get_brightness(), arv_camera_set_brightness(), and arv_camera_get_brightness_bounds() akin to what already exists for gain, exposure time, etc.

**Additional context**
See discussion on #582

PR #588, opened 2021-10-07
adding blacklevel support to arvcamera class

hackel6:
#586 
any comments / suggestions / corrections / complaints welcomed!

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ code coverage improvement is welcome.
 
 While most of the API is documented, Aravis documentation lacks some good
 tutorial about the many features if offers. But a good resource is the tests
-directory inside Aravis sources, where you will find a numbers of small samples
+directory inside Aravis sources, where you will find a set of small samples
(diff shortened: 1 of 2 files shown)


EmmanuelP:
> Our Basler cameras and our Imperx cameras both use "BlackLevelRaw." Shall I add an additional ArvCameraPrivate member for "uses black level raw"? Or i could add special cases for Vendor->Basler and Vendor->Imperx (new) and have all others fall down to "BlackLevel" default. Or I can just check for the existence of "BlackLevelRaw" register and if it's not supported, fall back down to "BlackLevel" default (basically this would be the same case as adding an additional ArvCameraPrivate, but inlining the check instead).

Please implement the first solution. I tend to avoid the second one lately, and the latest is less efficient, particularly for a setting that can be used in real time during the acquisition.

EmmanuelP:
Pushed to master.

Thanks Brian.

PR #590, opened 2021-10-13
camera: fallback to FrameBurstStart if FrameStart is not found

### LINKED ISSUE #589: Value not found in <Enumeration> 'TriggerSelector'. that is : How to use aravis for camera that only has trigger selector "FrameBurstStart"
**Describe the bug**
When run arv_camera_set_trigger, I get the bug:
```
Value not found in <Enumeration> 'TriggerSelector
```

I read the aravis document, and get this info: 

https://github.com/AravisProject/aravis/blob/aecb00ce97abdbac5b46ae868e787b5d1f500c13/src/arvcamera.c#L1236

```

 * The trigger set is "FrameStart". "AcquisitionStart" is used as a fallback if "FrameStart" is not present.
 * All other triggers are disabled. "TriggerActivation" is set to rising edge.
```

But when I check my camera, there is only one option in trigger selector.

the camera brand is hikrobotics, https://www.hikrobotics.com/.

Is there any way for me to use Aravis to capture image from it ? Or whether I have to consider other cameras ?

I find some code here, https://github.com/AravisProject/aravis/blob/aecb00ce97abdbac5b46ae868e787b5d1f500c13/src/arvcamera.c#L1263,
but I don't know how to edit it. 

please help me. 

### Update

I have found something here: https://docs.baslerweb.com/trigger-selector#frame-burst-start-acquisition-start-end-and-active-trigger

>  Frame Burst Start (= Acquisition Start), End, and Active Trigger#
> Info
> 
> The Frame Burst Start trigger and the Acquisition Start trigger are identical, only the names differ. The naming depends on your camera model. In the following, the term "Frame Burst Start trigger" is used to describe both types.

Could you add `Frame Burst Start` to Aravis ? Please. 

**To Reproduce**

**Expected behavior**
Capture image by software trigger from camera that only support trigger selector `frame burst start`.

**Camera description:**
 -  Manufacturer: hikrobotics
 - Model: MV-CA050-20GN
 - Interface: Gige

**Platform description:**
 - version : 0.8
 - OS: Windows 10
 - Hardware: x86_64

**Additional context**

Thank you !

EmmanuelP:
Fixes #589

### DIFF
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -1246,7 +1246,8 @@ arv_camera_set_trigger (ArvCamera *camera, const char *source, GError **error)
 {
 	GError *local_error = NULL;
 	gboolean has_frame_start = FALSE;
-	gboolean has_acquisition_start = FALSE;
+        gboolean has_frame_burst_start = FALSE; /* Hikrobot, Basler devices */
+	gboolean has_acquisition_start = FALSE; /* Smartek devices */
         const char **triggers = NULL;
         guint n_triggers = 0;
         unsigned int i;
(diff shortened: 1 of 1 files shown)


garfeng:
@EmmanuelP  thank you very much. The error gone !!!

Issue #540, opened 2021-07-25
-Wenum-conversion not understood by older compilers (gcc 9.3.0)


eudoxos:
https://github.com/AravisProject/aravis/blob/c139db0127a8a0d10e9c7cf87c81c9fd5f963030/meson.build#L34

I am using `x86_64-linux-gnu-gcc` cross-compiler under Linux, it is version `9.3.0`, and the `-Wenum-conversion` is not understood (it is understood by the native gcc 9.3.0, curiously):
```
x86_64-linux-gnu-gcc: error: unrecognized command line option ‘-Wenum-conversion’; did you mean ‘-Wno-conversion’?
```

EmmanuelP:
Hi,

The point of cc.get_supported_arguments is to filter the options to the one actually supported by the compiler. Which means -Wenum-conversion should not be passed to gcc if it is not supported. Are you trying to use meson cross compilation ?

eudoxos:
> Are you trying to use meson cross compilation ?

Yes, exactly. As described here: https://github.com/eudoxos/aravis/blob/winbuild/README.md#cross-compilation-for-windows

EmmanuelP:
Unless I have misunderstood crossroad, you are not using meson cross compilation support. I was asking because if meson cross compilation is used, there is a distinction in meson between the native compiler and the cross one, which could explain why the unsupported option is used during compilation.

https://mesonbuild.com/Cross-compilation.html

eudoxos:
`crossroad meson` will add the `--cross-file` to (native) `meson` according to the current platform; the wrapping is [here](https://git.tuxfamily.org/crossroad/crossroad.git/tree/src/in-crossroad.py#n334) and [this is the cross-file toolchain-w64.meson](https://git.tuxfamily.org/crossroad/crossroad.git/tree/platforms/meson/toolchain-w64.meson). So it should be running the cross-compiler.. Strange that it fails in this case. I will try to look closer what meson calls.

eudoxos:
After a closer look, I understand better now. The failure is when calling (native) `g-ir-scanner` which passes all flags for the cross-compiler to the native compiler. And as it happens, the cross-compiler does support `Wenum-conversion` (hence it does not get filtered out by `get_supported_arguments`) whereas the native gcc internally called by `g-ir-scanner` does not and fails. 

[...]

PR #591, opened 2021-10-15
ci: Update Conan package manager configuration for MSVC builds.

sh0:
Workflow updates:
* Adds new gst-plugins-base package to dependencies and enables gst-plugin.
* Removed pcre options since the default options for the packages are now upstreamed.
* Updates various package versions.

The viewer is still not enabled as building GTK on Windows has some bugs in Conan. The gobject introspection is very broken in Conan right now (similar problems as #540) so that is not going to be available soon.

I have tried vcpkg as an alternative to Conan, but I found it very troublesome and difficult to integrate in CI. Hence, I'd keep going with Conan.

### DIFF
--- a/.github/workflows/aravis-msvc.yml
+++ b/.github/workflows/aravis-msvc.yml
@@ -28,17 +28,17 @@ jobs:
       env:
         INPUT_CONANFILE: |
           [requires]
-          pcre/8.45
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Thanks a lot.

PR #596, opened 2021-10-29
Add GStreamer property to use asynchronous USB API

### LINKED ISSUE #595: Add GStreamer property for asynchronous USB mode
**Is your feature request related to a problem? Please describe.**
Currently, the USB mode (synchronous/asynchronous) cannot be set for the GStreamer element `aravissrc`.
Apparently, only the synchronous mode is used.

**Describe the solution you'd like**
An additional property of the GStreamer element. Suggestion: A boolean named `asynchronous_usb`

**Describe alternatives you've considered**
Setting the mode _via_ an environment variable might also work.

schaerfo:
Fixes #595

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ digits of the ID field.
 Aravis uses by default the synchronous libusb API. But it can be told to use the
 asynchronous API for better performances, especially on embedded platform like
 RapsberryPi or Nvidia Jetson boards. The function to use is
-`arv_camera_set_usb_mode()`.
+`arv_camera_uv_set_usb_mode()`.
 
 `arv-viewer` and `arv-camera-test` can also use the asynchronous API, if
 `usb-mode` option is set to `async`. When using the GStreamer plugin,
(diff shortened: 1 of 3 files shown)


EmmanuelP:
Hi @schaerfo 

Thanks for the merge request. I'd prefer if you convert the boolean property to an enum one, like what is done for exposure-auto, and rename usb-asynchronous to usb-mode. It would be more consistent with the tool parameters.

Do you think you can handle that ?

schaerfo:
Sure, that should be possible. The enum values would be synchronous, asynchronous and default, like in `ArvUvUsbMode`?

EmmanuelP:
> The enum values would be synchronous, asynchronous and default, like in ArvUvUsbMode?

Yes, `sync`,  `async` and `default` to be precise.

schaerfo:
> > The enum values would be synchronous, asynchronous and default, like in ArvUvUsbMode?
> 
> Yes, `sync`, `async` and `default` to be precise.

In this case, wouldn't it make sense to just re-use this enum?

EmmanuelP:
> In this case, wouldn't it make sense to just re-use this enum? Otherwise, we would have two separate enums with a similar purpose and the same values

Yes, same enum please.

Thanks.

EmmanuelP:
I'm going to push your changes to master. Thanks a lot.

PR #600, opened 2021-11-16
arv_camera_set_frame_rate: AcquisitionFrameRateEnable is not available on some Basler camera models

thiesmoeller:
The feature 'AcquisitionFrameRateEnable' is not available on Basler dart family models 
and has to be used with prior check for availability.
( any value in feature 'AcquisitionFrameRate'  works as a clipping value for the frame rate )


EmmanuelP:
Thanks.

PR #620, opened 2022-01-11
Add FreeBSD support

rozhuk-im:
Add FreeBSD support and probably fix other OSes to not show network interfaces without BROADCAST flag.

### DIFF
--- a/src/arvnetwork.c
+++ b/src/arvnetwork.c
@@ -365,9 +365,18 @@ arv_enumerate_network_interfaces (void)
 			a->addr = arv_memdup (ifap_iter->ifa_addr, sizeof(struct sockaddr));
 			if (ifap_iter->ifa_netmask)
 				a->netmask = arv_memdup (ifap_iter->ifa_netmask, sizeof(struct sockaddr));
-#if defined(__APPLE__) && defined(__MACH__)
-			if (ifap_iter->ifa_broadaddr)
-				a->broadaddr = arv_memdup(ifap_iter->ifa_broadaddr, sizeof(struct sockaddr));
+#if (defined(__APPLE__) && defined(__MACH__)) || defined(BSD)
+			if (ifap_iter->ifa_broadaddr &&
+			    ifap_iter->ifa_broadaddr->sa_len != 0) {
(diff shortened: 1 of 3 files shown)


rozhuk-im:
Have no idea why Linux and MSVC fail, looks like some temporary errors in test system.

rozhuk-im:
After I disable ```if ((ifap_iter->ifa_flags & IFF_BROADCAST) == 0)``` test is passed, but IMHO this is wrong, and interfaces without broadcast flag must not be used.

src/arvnetwork.c
```diff
@@ -354,6 +354,10 @@ arv_enumerate_network_interfaces (void)
 		return NULL;
 
 	for (ifap_iter = ifap; ifap_iter != NULL; ifap_iter = ifap_iter->ifa_next) {
+//#ifdef IFF_BROADCAST
+//		if ((ifap_iter->ifa_flags & IFF_BROADCAST) == 0)
+//			continue; /* Skip interfaces without broadcast. */
+//#endif
```
EmmanuelP, line 360: Please remove this part of commented code.

EmmanuelP (changes requested):
Thanks for the patch. Once the commented code is removed, it will be ready to merge.

EmmanuelP:
> After I disable `if ((ifap_iter->ifa_flags & IFF_BROADCAST) == 0)` test is passed, but IMHO this is wrong, and interfaces without broadcast flag must not be used.

That is not really related to the main goal of this pull request, so the pull request can be merged without this part. But feel free to investigate why this test makes the fakegv test fail on linux and macOS in another pull request.

src/arvnetworkprivate.h
```diff
@@ -38,16 +39,22 @@
...
 #endif
 
-#if !defined(__APPLE__) && !defined(G_OS_WIN32)
+#ifdef __linux__
 #include <linux/ip.h>
 #endif
 
+#if defined(__APPLE__) || defined(BSD)
+#include <netinet/in.h>
+#include <netinet/ip.h>
```
EmmanuelP, line 48: Redundant include.

EmmanuelP (changes requested):

EmmanuelP:
Thanks a lot.

PR #623, opened 2022-01-31
Add dependencies for Fedora

danofsteel32:
Needed to install these packages in order to support all of the build options on Fedora 34/35.


EmmanuelP:
Thanks.

PR #642, opened 2022-03-02
fix PACKETRESEND size in header for extended ID mode

apramskiy:
arv_gvcp_packet_new_packet_resend_cmd() allocates a correct memory size for both
standard and extended ID modes. But a size value inside header is not correct
in case of an extended ID mode. Value is always like for standard ID mode.


EmmanuelP:
Thanks

PR #644, opened 2022-03-07
Update path to aravis icon: fix AppStream warning.

cmarmo:
This pull request update the path to the aravis icon.
While github in general redirects _master_ links to _main_, AppStream doesn't know about that, then the icon is not retrieved.

Found in packaging for Debian.

Thanks for considering it.


EmmanuelP:
Thanks Chiara.

PR #653, opened 2022-03-30
uv: register to libusb device-left event for control-lost signal emission

EmmanuelP:


romainreignier:
Tested with Allied Vision Alvium 1800U in Async mode.
The  `arv-camera-test` tool now detects the control-lost properly.

PR #669, opened 2022-05-09
camera: remove duplicate of arv_camera_clear_triggers

romainreignier:
In the process of supporting Imperx cameras, I will need to make the usage of TriggerSelector optional.
But it appears that the implementation of `arv_camera_disable_all_triggers` introduced 8 months ago is the same as the `arv_camera_clear_triggers` that was already there.


EmmanuelP:
Thanks.

PR #679, opened 2022-06-02
arvgcport: add a list of devices using legacy infos

romainreignier:
Some devices advertise a schema version 1.1.0 but still use legacy legacy memory access methods.

This PR introduces a list of devices that is checked to chose the right method.

This is only a draft for now, any comments are welcomed.

### DIFF
--- a/src/arvgcport.c
+++ b/src/arvgcport.c
@@ -37,8 +37,19 @@
 #include <memory.h>
 
 typedef struct {
+	const char *vendor_selection;
+	const char *model_selection;
+} ArvGvLegacyInfos;
+
+static ArvGvLegacyInfos arv_gc_port_legacy_infos[] = {
+   { .vendor_selection = "Imperx", .model_selection = "IpxGEVCamera"},
+};
+
+typedef struct {
 	ArvGcPropertyNode *chunk_id;
 	ArvGcPropertyNode *event_id;
+	gboolean has_done_legacy_check;
+	gboolean has_legacy_infos;
 } ArvGcPortPrivate;
 
 struct _ArvGcPort {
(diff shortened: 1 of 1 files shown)


src/arvgcport.c
```diff
@@ -248,8 +252,32 @@ arv_gc_port_new (void)
...
+	ArvGcRegisterDescriptionNode *register_description;
+	const char *vendor_name;
+	const char *model_name;
+	int i;
+
 	gc_port->priv = arv_gc_port_get_instance_private (gc_port);
 	gc_port->priv->chunk_id = 0;
+	gc_port->priv->has_legacy_infos = FALSE;
+
+	document = arv_dom_node_get_owner_document (ARV_DOM_NODE (gc_port));
```
romainreignier, line 265: This line fail with error:
> arv_dom_document_get_document_element: assertion 'ARV_IS_DOM_DOCUMENT (self)' failed

I guess that the document is not yet initialized.
Where can I move this code?

In the first call of `_use_legacy_endianness_mechanism`?

romainreignier:
This draft is not working, the document check is not working, probably done too early, see my comment above.

romainreignier:
With this version, doing the check in `_use_legacy_endianness_mechanism` and with one more variable to know if the check has been done works:
https://github.com/romainreignier/aravis/commit/d0c2b2761823f196e61fcfc57098056d0ad002c9

src/arvgcport.c
EmmanuelP: You can not access the document element in the init function, the node is not yet attached to the XML tree.

This code should be called during the first call of _use_legacy_endianness_mechanism.

romainreignier: Ok, I see, thank you.
I have then pushed a new version calling this in the first call of `_use_legacy_endianness_mechanism`.

PR #680, opened 2022-06-03
arvbuffer: fix declaration of user_data_destroy_func

tpwrules:
This allows a callback to not be passed from an FFI, e.g. PyGObject.


EmmanuelP:
Thanks.

Issue #528, opened 2021-07-15
Aravissrc times out when frame acquisition is not at a fixed frame rate


PauloFavero:
**Describe the bug**

Not able to use aravissrc as a source when triggering the camera by hardware. The pipeline does not send the buffer downstream. The goal is to generate snapshots when triggering the camera from the hardware. 

**To Reproduce**
1 - Set the camera to generate frames on demand when triggered by hardware events:
```
arv-tool-0.8 -n 'BALLUFF GmbH-BVS CA-GX0-0032AC-GX301282' control AcquisitionMode='Continuous'
arv-tool-0.8 -n 'BALLUFF GmbH-BVS CA-GX0-0032AC-GX301282' control ExposureTime='200000'
arv-tool-0.8 -n 'BALLUFF GmbH-BVS CA-GX0-0032AC-GX301282' control TriggerSelector='FrameStart'
arv-tool-0.8 -n 'BALLUFF GmbH-BVS CA-GX0-0032AC-GX301282' control TriggerMode='On'
arv-tool-0.8 -n 'BALLUFF GmbH-BVS CA-GX0-0032AC-GX301282' control TriggerActivation='FallingEdge'
arv-tool-0.8 -n 'BALLUFF GmbH-BVS CA-GX0-0032AC-GX301282' control TriggerSource='Line4'
```

2 - Launch gstd client:

```
gstd -p 5000 -n 2
```

3 - In another terminal, create and play the pipeline:

```
 gst-client pipeline_create test aravissrc name=balluff num-buffers=10 \
 ! video/x-bayer, format=rggb, width=1920, height=1080, framerate=0/1, pixel-aspect-ratio=1/1 ! bayer2rgb \
 ! video/x-raw, format=RGBA ! videoconvert ! jpegenc ! multifilesink async=false location=test.jpeg
 
 gst-client debug_enable true
 gst-client debug_color true
 gst-client debug_threshold **:6

 gst-client element_set test balluff camera-name "BALLUFF GmbH-BVS CA-GX0-0032AC-GX301282"

 sleep 5
 gst-client pipeline_play test

 sleep 3
 gst-client pipeline_play test
```

4 - Trigger the camera.

In our setup, we have a camera with pin 4 connected to a laser sensor. When the laser is interrupted, we have a failing edge that triggers the camera to capture an image. This behavior works very well when running the wxPropView provided by the Balluff SDK.

**Expected behavior**
[...]

PR #686, opened 2022-06-28
gst: Add GstQueryLatency handling

TIS-Edgar:
Define maximum latency and GST_CLOCK_TIME_NONE to prevent timeouts.

Potential fix for #528.

All gstreamer sources must be synchronized with the gstreamer clock.
Since aravissrc is a live source it can't give reliable increasing timestamps and instead works with durations.
Ideally that would mean first timestamp (0) + duration of frame. New timestamp + duration of frame, etc.
Since we cannot know what the duration of a frame is (some come faster, some slower, thus never being 100% accurate) gstreamer wants to know a latency range. 
This patch defined the latency as min=framerate and max=GST_CLOCK_TIME_NONE.
That basically says expect the next buffer between average duration and any given time in the future after that,
users more than enough time to trigger, extended exposure times, etc.

### DIFF
--- a/gst/gstaravis.c
+++ b/gst/gstaravis.c
@@ -280,17 +280,16 @@ gst_aravis_set_caps (GstBaseSrc *src, GstCaps *caps)
 	}
 
 	if (!error && frame_rate != NULL) {
-		double dbl_frame_rate;
 
-		dbl_frame_rate = (double) gst_value_get_fraction_numerator (frame_rate) /
+		gst_aravis->frame_rate = (double) gst_value_get_fraction_numerator (frame_rate) /
 			(double) gst_value_get_fraction_denominator (frame_rate);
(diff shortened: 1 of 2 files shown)


TIS-Edgar:
I have no clue why the Aravis-MSVC check fails with

    source_subfolder\meson.build:291:0: ERROR: Dependency 'glib-2.0' is required but not found.

EmmanuelP:
> I have no clue why the Aravis-MSVC check fails with
> 
> ```
> source_subfolder\meson.build:291:0: ERROR: Dependency 'glib-2.0' is required but not found.
> ```

Don't worry about that, this pipeline is failing for a while now, probably due to a change in the build environment.

EmmanuelP:
Thanks Edgar

Issue #688, opened 2022-07-07
Feature Request: CLI to configure IP settings


romainreignier:
**Is your feature request related to a problem? Please describe.**

As mentioned here: https://github.com/AravisProject/aravis/pull/681#pullrequestreview-994512851
It would be nice to have a command line tool to configure the IP settings of a camera.

So I open an issue to track this idea.

**Describe the solution you'd like**
I might try to implement it.
Should we create a new tool or add this functionality to `arv-tool`?

The base options should be to:
* Set Mode (LLA, HCP or persistent IP)
* Set IP
* Set Submask
* Set default gateway

**Describe alternatives you've considered**
I am not aware of any open-source tool allowing to change these settings on Linux from the command line.
If there are some, I will be glad to know them.

EmmanuelP:
I think it is better to add this to arv-tool.

romainreignier:
Ok, I agree with that.
About the "API", can we do something like the control mode?
```
arv-tool-0.8 network mode=PersistentIP
arv-tool-0.8 network ip=192.168.0.1
```

romainreignier:
`arv-tool` uses `ArvDevice` and not an `ArvCamera` object.
But I will need to convert IP addresses from string to `GInetAddress` like it is done in `arv_camera`:

https://github.com/AravisProject/aravis/blob/af9a095479cce8cff22a538124732eb20bb3ed5f/src/arvcamera.c#L3049-L3082

Should I duplicate this code in `arvtool.c`, use an `ArvCamera` with `arv_camera_new_with_device()` or move this helper to `arvdevice.c`?

Actually, to use the function `arv_camera_gv_set_persistent_ip_from_string()` I need to have all three settings (ip, submask and gateway) at the same time, so the API will be:
```
arv-tool-0.8 network ip=192.168.0.1,255.255.255.0,192.168.0.254
```

Is this ok, or should we prefer to set only one setting at a time?

EmmanuelP:
Please move the helper to arvdevice.

It will be better to be able to set ip mask and gateway independently.  You can modify the helper to ignore the arguments if they are NULL.

PR #690, opened 2022-07-07
Feature/add network settings arvtool

romainreignier:
As mentioned here https://github.com/AravisProject/aravis/issues/688 I have added a `network` command to `arvtool`.

I have several remarks:

* Please check the error usage, I am not sure on how to handle errors correctly
* Please check memory management to check if there are memory leaks with glib interactions.
* I am not sure on how to document the new arvtool API
* Should the `network` command support setting several parameters on one line like the `control` command?
* I have a version of the `ip` sub-command with ip, mask and gateway separated with a coma. is this acceptable?
* The `ip` only sub-command prints all three settings (ip, mask, gateway) while `mask` and `gateway` only print their setting. Should I change that?
* `arv_gv_device_get_persistent_ip` always return the IP address but in some cases, I don't use it, is it ok or should I make it optional like the `mask` and `gateway`?
* I had to change this line (and related):
    https://github.com/AravisProject/aravis/blob/af9a095479cce8cff22a538124732eb20bb3ed5f/src/arvgvdevice.c#L898
    By this one:
    ```c++
    be_value = ((guint32)ip_bytes[3] << 24) | (ip_bytes[2] << 16) | (ip_bytes[1] << 8) | ip_bytes[0];
    ```
    Otherwise the set IP was `10.0.168.192` instead of `192.168.0.10`.

### DIFF
--- a/src/arvgvdevice.c
+++ b/src/arvgvdevice.c
@@ -1015,9 +1015,26 @@ arv_gv_device_set_persistent_ip_from_string (ArvGvDevice *gv_device,
 ArvGvIpConfigurationMode
 arv_gv_device_get_ip_configuration_mode (ArvGvDevice *gv_device, GError **error)
 {
+	gboolean dhcp_enabled;
+	gboolean persistent_ip_enabled;
+
 	g_return_val_if_fail (ARV_IS_GV_DEVICE (gv_device), 0);
 
-	return arv_device_get_integer_feature_value (ARV_DEVICE (gv_device), "GevIPConfigurationStatus", error);
+	if (arv_device_is_feature_available (ARV_DEVICE (gv_device), "GevIPConfigurationStatus", error))
+		return arv_device_get_integer_feature_value (ARV_DEVICE (gv_device), "GevIPConfigurationStatus", error);
+
+	dhcp_enabled = arv_device_get_boolean_feature_value( ARV_DEVICE (gv_device), "GevCurrentIPConfigurationDHCP", error);
+	if (*error != NULL)
+		return ARV_GV_IP_CONFIGURATION_MODE_NONE;
+	persistent_ip_enabled = arv_device_get_boolean_feature_value( ARV_DEVICE (gv_device), "GevCurrentIPConfigurationPersistentIP", error);
+	if (*error != NULL)
+		return ARV_GV_IP_CONFIGURATION_MODE_NONE;
+
+	if (dhcp_enabled && !persistent_ip_enabled)
+		return ARV_GV_IP_CONFIGURATION_MODE_DHCP;
+	if (!dhcp_enabled && persistent_ip_enabled)
+		return ARV_GV_IP_CONFIGURATION_MODE_PERSISTENT_IP;
+	return ARV_GV_IP_CONFIGURATION_MODE_LLA;
 }
 
 
--- a/src/arvgvdevice.c
+++ b/src/arvgvdevice.c
@@ -915,6 +915,59 @@ arv_gv_device_set_persistent_ip (ArvGvDevice *gv_device,
 	arv_gv_device_set_ip_configuration_mode (gv_device, ARV_GV_IP_CONFIGURATION_MODE_PERSISTENT_IP, NULL);
 }
 
+
+/**
+ * arv_gv_device_set_persistent_ip_from_string:
+ * @gv_device: a #ArvGvDevice
+ * @ip: IPv4 address in string format
+ * @mask: netmask in string format
+ * @gateway: Gateway IPv4 address in string format
+ * @error: a #GError placeholder, %NULL to ignore
+ *
+ * Sets the persistent IP address to device.
+ *
+ * Since: 0.8.22
+ */
+
+void
+arv_gv_device_set_persistent_ip_from_string (ArvGvDevice *gv_device,
+                                             const char *ip, const char *mask, const char *gateway,
+                                             GError **error)
+{
+	GError *local_error = NULL;
+	GInetAddress *ip_gi = NULL;
+	GInetAddressMask *mask_gi = NULL;
+	GInetAddress *gateway_gi = NULL;
+
+	g_return_if_fail (ARV_IS_GV_DEVICE (gv_device));
+
+	ip_gi = g_inet_address_new_from_string (ip);
+	mask_gi = g_inet_address_mask_new_from_string (mask, NULL);
+	gateway_gi = g_inet_address_new_from_string (gateway);
+
+	if (ip_gi == NULL) {
+		local_error = g_error_new (ARV_DEVICE_ERROR, ARV_DEVICE_ERROR_INVALID_PARAMETER,
+                                           "IP address could not be parsed: \"%s\"", ip);
+	}else if (mask_gi == NULL) {
+		local_error = g_error_new (ARV_DEVICE_ERROR, ARV_DEVICE_ERROR_INVALID_PARAMETER,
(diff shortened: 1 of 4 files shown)


src/arvgvdevice.c
```diff
@@ -842,9 +842,9 @@ arv_gv_device_get_persistent_ip (ArvGvDevice *gv_device,
 /**
  * arv_gv_device_set_persistent_ip:
  * @gv_device: a #ArvGvDevice
- * @ip: IPv4 address
- * @mask: Netmask
- * @gateway: Gateway IPv4 address
```
EmmanuelP, line 847: Please use the (optional) annotation here:
```
* @ip: (optional): IPv4 address
```

src/arvgvdevice.c
```diff
@@ -858,61 +858,136 @@ arv_gv_device_set_persistent_ip (ArvGvDevice *gv_device,
...
-		be_value = ~(~(guint32)0 >> mask_length);
+
+/**
+ * arv_gv_device_set_persistent_ip_from_string:
+ * @gv_device: a #ArvGvDevice
+ * @ip: IPv4 address in string format, %NULL to ignore
+ * @mask: netmask in string format, %NULL to ignore
+ * @gateway: Gateway IPv4 address in string format, %NULL to ignore
+ * @error: a #GError placeholder, %NULL to ignore
+ *
```
EmmanuelP, line 938: (optional)

src/arvgvdevice.c
```diff
@@ -858,61 +858,136 @@ arv_gv_device_set_persistent_ip (ArvGvDevice *gv_device,
...
+		be_value = ((guint32)gateway_bytes[3] << 24) | (gateway_bytes[2] << 16) | (gateway_bytes[1] << 8) | gateway_bytes[0];
+		gateway_int = GUINT32_FROM_BE(be_value);
+		arv_device_set_integer_feature_value (ARV_DEVICE (gv_device), "GevPersistentDefaultGateway", gateway_int, NULL);
 	}
 
-	ip_bytes = g_inet_address_to_bytes (ip);
-	mask_length = g_inet_address_mask_get_length (mask);
-	mask_bytes = g_inet_address_to_bytes (g_inet_address_mask_get_address (mask));
-	gateway_bytes = g_inet_address_to_bytes (gateway);
+	arv_gv_device_set_ip_configuration_mode (gv_device, ARV_GV_IP_CONFIGURATION_MODE_PERSISTENT_IP, NULL);
```
EmmanuelP, line 927: Error should not be ignored

src/arvgvdevice.c
```diff
@@ -858,61 +858,136 @@ arv_gv_device_set_persistent_ip (ArvGvDevice *gv_device,
...
-	arv_device_set_integer_feature_value (ARV_DEVICE (gv_device), "GevPersistentIPAddress", ip_int, NULL);
-	arv_device_set_integer_feature_value (ARV_DEVICE (gv_device), "GevPersistentSubnetMask", mask_int, NULL);
-	arv_device_set_integer_feature_value (ARV_DEVICE (gv_device), "GevPersistentDefaultGateway", gateway_int, NULL);
-	arv_gv_device_set_ip_configuration_mode (gv_device, ARV_GV_IP_CONFIGURATION_MODE_PERSISTENT_IP, NULL);
+	if (ip_gi != NULL)
+		g_object_unref (ip_gi);
+	if (mask_gi != NULL)
+		g_object_unref (mask_gi);
+	if (gateway_gi != NULL)
+		g_object_unref (gateway_gi);
```
EmmanuelP, line 990: Use:

```
	g_clear_object (&ip_gi);
	g_clear_object (&mask_gi);
	g_clear_object (&gateway_gi);
```

instead.

src/arvtool.c
```diff
@@ -394,6 +399,164 @@ arv_tool_control (int argc, char **argv, ArvDevice *device)
         }
 }
 
+static void
+arv_tool_network (int argc, char **argv, ArvDevice *device)
+{
+        if (!ARV_IS_GV_DEVICE (device))
+                printf ("This is not a GV device\n");
```
EmmanuelP, line 406: Don't use an if else structure, but return early if the device is not a GigE one.

src/arvtool.c
```diff
@@ -394,6 +399,164 @@ arv_tool_control (int argc, char **argv, ArvDevice *device)
...
+        else if (argv[2] != NULL) {
```
EmmanuelP, line 407: It would be nice if by default `network` alone displays the network configuration

EmmanuelP, line 407: I would prefer if there was a loop here, like in the 'control' command.

And the ip subcommand changed to only take an ip address.

EmmanuelP (changes requested):
Looks mostly fine. I did not spot any memory leak.

src/arvtool.c
romainreignier: I agree with that. After some more thoughts I think this is a better solution.

romainreignier: For now, the `ip` shows the persistent IP configuration, which is different from the **current** configuration. Maybe we should show both?

EmmanuelP: Yes

src/arvgvdevice.c
EmmanuelP: Try: `@ip: (out) (optional): IPv4 address

romainreignier: Ok, thanks but in the case of `arv_gv_device_set_persistent_ip` this is an input parameter.

EmmanuelP: Ah. (nullable) should work then.

romainreignier: Should I keep the `, %NULL to ignore` at the end of the line?

EmmanuelP: Here is the annotation documentation: https://gi.readthedocs.io/en/latest/annotations/giannotations.html

And indeed (nullable) should be used for (in) parameters.

EmmanuelP: Don't keep `%NULL% to ignore`, it is redundant with the introspection annotation. gi-docgen should translate the annotation into a full text explanation.

src/arvtool.c
```diff
@@ -394,6 +398,183 @@ arv_tool_control (int argc, char **argv, ArvDevice *device)
...
+                int i;
+
+                for (i = 2; i < argc; i++) {
+                        GError *error = NULL;
+                        char **tokens;
+
+                        tokens = g_strsplit (argv[i], "=", 2);
+                        if (g_strcmp0 (tokens[0], "mode") == 0) {
+                                if (tokens[1] != NULL) {
+                                        ArvGvIpConfigurationMode mode;
```
EmmanuelP, line 537:
```sh
Dans la fonction « arv_tool_network »,
    mis en ligne depuis « arv_tool_execute_command » à ../src/arvtool.c:639:17:
../src/arvtool.c:545:41: attention: « mode » pourrait être utilisé sans être initialisé [-Wmaybe-uninitialized]
  545 |                                         arv_gv_device_set_ip_configuration_mode (gv_device, mode, &error);
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/arvtool.c: Dans la fonction « arv_tool_execute_command »:
../src/arvtool.c:537:66: note: « mode » a été déclaré ici
  537 |                                         ArvGvIpConfigurationMode mode;
      |                                                                  ^~~~
```

EmmanuelP (changes requested):

PR #699, opened 2022-07-23
Implemented AcquisitionFrameCount on Fake-camera

LorenzoBalducci96:
The first frame is immediately taken, but maybe is correct to wait the framePeriod also on the first frame?


.gitignore
```diff
@@ -3,3 +3,251 @@ build/
...
+# Paket dependency manager
+.paket/paket.exe
+paket-files/
+
+# FAKE - F# Make
+.fake/
+
+# JetBrains Rider
+.idea/
+*.sln.iml
```
EmmanuelP, line 253: Please don't include unrelated changes.

src/arvfakecamera.c
```diff
@@ -1074,6 +1105,7 @@ arv_fake_camera_init (ArvFakeCamera *fake_camera)
 
 	fake_camera->priv->trigger_frequency = 25.0;
 	fake_camera->priv->frame_id = 65400; /* Trigger circular counter bugs sooner */
+	fake_camera->priv->frameSeqCount = 65535;//max possible value by default for acquisition count
```
EmmanuelP, line 1108: don't use C++ style comments

src/arvgvfakecamera.c
```diff
@@ -265,7 +265,6 @@ _thread (void *user_data)
```
EmmanuelP, line 265: It looks like there is no real change in this file.

src/arvfakecamera.c
```diff
@@ -906,6 +925,18 @@ arv_fake_camera_get_control_channel_privilege (ArvFakeCamera *camera)
...
+	return _get_register (camera, ARV_FAKE_CAMERA_REGISTER_ACQUISITION_FRAME_PERIOD_US);
+}
+
+int
+arv_fake_camera_get_acquisition_frame_count (ArvFakeCamera *camera)
+{
+	return _get_register (camera, ARV_FAKE_CAMERA_ACQUISITION_FRAME_COUNT);
+}
+
 void
```
EmmanuelP, line 940: These functions are never used.

src/arvfakecamera.c
```diff
@@ -72,6 +72,7 @@ typedef struct {
 
 	guint32 frame_id;
 	double trigger_frequency;
+        int frameSeqCount;
```
EmmanuelP, line 75: Should be multi_frame_count instead of frameSeqCount.

src/arv-fake-camera.xml
```diff
@@ -336,6 +336,21 @@
...
 
+  <Integer Name="AcquisitionFrameCount" NameSpace="Custom">
+    <pValue>AcquisitionFrameCountRegister</pValue>
+    <Min>1</Min>
+    <Max>65535</Max>
+  </Integer>
+
+  <IntReg Name="AcquisitionFrameCountRegister" NameSpace="Custom">
+    <Address>0x310</Address>
+    <Length>4</Length>
```
EmmanuelP, line 347: Use a 64bit integer, 65535 seems to small.

EmmanuelP (changes requested):
Hi,

Thanks for your work on this feature. It could go in, but there is several things to fix first.

In addition to the other comments, there is also this 2 points:

FrameCount should only taken into account if AcquisitionMode is MultiFrame.

I'm not sure, but it looks like the counter is not reset at every AcquisitionStart.

EmmanuelP:
Fake camera should stick to the behavior describe in the Genicam SNFC document: https://www.emva.org/wp-content/uploads/GenICam_SFNC_v2_7.pdf

LorenzoBalducci96:
Hi thanks for the interest.

From the Genicam SNFC document I read: "AcquisitionFrameRate controls the rate at which the Frames are captured when TriggerMode is Off.".
So it seems that this parameter must be taken into account when not using triggered mode acquisition. (so I guess it defines the framerate when acquisitionMode is continous). So it should be correct to not take into account the framerate when sending multiFrames in triggered mode and send frames immediately as soon as are available in the fake-camera?.
However in basler aca1920-50gm/c you can set the AcquisitionFrameRateAbs parameter when acquiring multiFrame in triggered mode and it will define the framerate of the frames acquired after a trigger. 
Maybe basler cameras are slightly out of specification?

"the counter is not reset at every AcquisitionStart"
in arv_fake_camera_check_and_acknowledge_software_trigger when the trigger is acknoledged the frameCount is reset to 1 and is increased every new frame. Since the frameCount is smaller than AcquisitionFrameCount register the arv_fake_camera_check_and_acknowledge_software_trigger still returns true meaning that a new frame must be taken (the trigger was already acknoledged at first frame). Once the acquisitionFrameCount is reached, the frameCount is set to 65535 in order to be bigger than any other acquisitionFrameCount that could be set before the next trigger causes the reset of the frameCount.
Do you propose a better solution?

every other notes are for shure reasonable for the next commit.

EmmanuelP:
> From the Genicam SNFC document I read: "AcquisitionFrameRate controls the rate at which the Frames are captured when TriggerMode is Off.". So it seems that this parameter must be taken into account when not using triggered mode acquisition. (so I guess it defines the framerate when acquisitionMode is continous). So it should be correct to not take into account the framerate when sending multiFrames in triggered mode and send frames immediately as soon as are available in the fake-camera?. 

Yes.

> However in basler aca1920-50gm/c you can set the AcquisitionFrameRateAbs parameter when acquiring multiFrame in triggered mode and it will define the framerate of the frames acquired after a trigger. Maybe basler cameras are slightly out of specification?

Most camera manufacturers don't strictly stick to the SNFC, mostly because they started the work on their firmware and SDK before the SNFC even exists.

> "the counter is not reset at every AcquisitionStart" in arv_fake_camera_check_and_acknowledge_software_trigger when the trigger is acknoledged the frameCount is reset to 1 and is increased every new frame. Since the frameCount is smaller than AcquisitionFrameCount register the arv_fake_camera_check_and_acknowledge_software_trigger still returns true meaning that a new frame must be taken (the trigger was already acknoledged at first frame). Once the acquisitionFrameCount is reached, the frameCount is set to 65535 in order to be bigger than any other acquisitionFrameCount that could be set before the next trigger causes the reset of the frameCount. Do you propose a better solution?

There is no reason to couple AcquisitionFrameCount and TriggerSoftware, they are unrelated. AcquisitionFrameCount is here to set the number of acuiqred frames when AcuisitionMode is set to MultiFrame, wether the camera is free running or triggered.

PR #701, opened 2022-08-05
camera: Add workaround to set exposure time and check if available on Imperx Cheetah

romainreignier:
Feature ExposureTime is not available if ExposureMode is not set to Timed
so check for ExposureMode in arv_camera_is_exposure_time_available.

I don't know for the other series so change the behavior on series instead of vendor.


EmmanuelP:
Thanks.

PR #712, opened 2022-09-01
ci: enable gst-plugin in msvc pipeline

Fixstars-momoko:
in msvc pipeline

1. disabled strawberry perl and added pkg-config on conanfile
2. enabled gst-plugin

---

1. somehow the latest meason does not support pkg-config bundled with strawberry perl. So removed directory under `C:\Strawberry\perl`

Runner image of [Microsoft Windows Server 2019 Datacenter](https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md) has Perl 5.32.1 as a default and it is located under `C:\Strawberry\perl`

image

2.  pkg-config installed as dependencies with conan now can find glib-2.0 and gst-plugins-base is successfully installed

### DIFF
--- a/.github/workflows/aravis-msvc.yml
+++ b/.github/workflows/aravis-msvc.yml
@@ -22,6 +22,9 @@ jobs:
     - name: pip
       run: |
         pip install conan
+    - name: disable-perl
+      run: |
+        rm -r C:\Strawberry\perl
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Thanks a lot @Fixstars-momoko

Issue #453, opened 2020-12-01
python: stream callback crash with segfault.


bpinsard:
**Describe the bug**
I tried setting a stream callback as a python function, following examples in C.

```
def stream_cb(user_data, type, buffer):
    if type == Aravis.StreamCallbackType.INIT:
        if not Aravis.make_thread_realtime(10) and \
            not Aravis.make_thread_high_priority(-10):
            print("Failed to make stream thread high priority")
``` 
and passing it to the stream as
```
stream=cam.create_stream(stream_cb, None)
```

However, my code crashes with segfault.
I suppose this is due to the stream thread accessing to some of the main thread memory but I am not sure how to properly separate these.
Thanks in advance for your help.

**To Reproduce**
see above

**Camera description:**
 - Manufacturer Fake
 - Model Fake
 - Interface [e.g. USB3] Ethernet

**Platform description:**
 - Aravis version 0.8.5
 - OS: Ubuntu 20
 - Hardware x86_64

EmmanuelP:
Could you attach a crash trace ?

bpinsard:
The crash produce files in `cysignals_crash_logs/` such as that one:
[crash_3rb4oas6.log](https://github.com/AravisProject/aravis/files/5629568/crash_3rb4oas6.log)
Is it what you meant?

EmmanuelP:
> Is it what you meant?

Yes, thanks.

Unfortunately, it is mostly gibberish for me, I'm not used to debug crashes from python code.

If I'm not mistaken, make_thread_realtime and make_thread_high_priority should be thread safe.

Did you try to test with just make_thread_realtime, then just make_thread_high_priority ?

EmmanuelP:
Oups, I did not meant to close this issue.

bpinsard:
I even tried a function that does nothing (just return) and it crashes the same, so it is likely due to python.

bpinsard:
If you want to reproduce here is what to run in python:
```
import gi 
gi.require_version('Aravis', '0.8') 
from gi.repository import Aravis 
uid="Aravis-Fake-GV01" 
cam = Aravis.Camera.new(uid) 
def cb(user_data, cb_type, buffer): 
    print(user_data, cb_type, buffer) 
stream=cam.create_stream(cb, None) 
payload = cam.get_payload() 
stream.push_buffer(Aravis.Buffer.new_allocate(payload))                                                                      
```
Which should produce the following output and crash, it crashes when pushing a buffer.

```
[...]

PR #714, opened 2022-09-09
Fix crash when using stream callback in Python

lostcontrol:
This PR should fix the crash seen in #453. The crash was caused by `ArvStreamCallback` only living for the time of the call to `arv_camera_create_stream` because of `(scope call)`. Using `(scope notified)` allows us to control the lifetime of the callback.

Also added a small Python sample showing how to use stream callback.

:warning: GObject & GLib noob here, careful review required.

### DIFF
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -137,7 +137,7 @@ enum
 };
 
 /**
- * arv_camera_create_stream:
+ * arv_camera_create_stream: (skip)
  * @camera: a #ArvCamera
  * @callback: (scope call) (allow-none): a frame processing callback
  * @user_data: (closure) (allow-none): user data for @callback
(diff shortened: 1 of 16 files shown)


lostcontrol:
Applying `clang-format` to the files results in many unrelated changes. It seems also like there is a mix of tabs and spaces already.

I didn't write any tests. Should I extend `tests/fake.py` to also use a callback?

EmmanuelP:
> Applying `clang-format` to the files results in many unrelated changes. It seems also like there is a mix of tabs and spaces already.

Don't bother about clang-format. I should remove it from the repository.

> I didn't write any tests. Should I extend `tests/fake.py` to also use a callback?

That would be nice, yes.

src/arvdevice.c
```diff
@@ -82,10 +82,32 @@ G_DEFINE_ABSTRACT_TYPE_WITH_CODE (ArvDevice, arv_device, G_TYPE_OBJECT,
...
+ * @user_data: (allow-none) (closure): user data for @callback
+ * @destroy: a #GDestroyNotify placeholder, %NULL to ignore
+ * @error: a #GError placeholder, %NULL to ignore
+ *
+ * Creates a new #ArvStream for video stream handling. See
+ * @ArvStreamCallback for details regarding the callback function.
+ *
+ * Return value: (transfer full): a new #ArvStream.
+ *
+ * Since: 0.2.0
```
EmmanuelP, line 102: Since 0.8.23

EmmanuelP (changes requested):
Looks fine.

The ArvDevice::create_stream signature change is an API break, but adding a new create_stream_full would still be an ABI break. ArvDevice virtual methods are not really meant to be public, so your change is fine.

src/arvcamera.c
```diff
@@ -153,12 +153,34 @@ enum
...
+ * @user_data: (closure) (allow-none): user data for @callback
+ * @destroy: a #GDestroyNotify placeholder, %NULL to ignore
+ * @error: a #GError placeholder, %NULL to ignore
+ *
+ * Creates a new [class@ArvStream] for video stream reception. See
+ * [callback@ArvStreamCallback] for details regarding the callback function.
+ *
+ * Returns: (transfer full): a new [class@ArvStream], to be freed after use with [method@GObject.Object.unref].
+ *
+ * Since: 0.2.0
```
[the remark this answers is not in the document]
lostcontrol: Done

src/arvdevice.c
lostcontrol: Done

lostcontrol:
I also quickly tested with a Lucid TRI064S-M. Seems to work fine.

EmmanuelP:
Thanks a lot Cyril. Everything looks fine now.

I'm still a bit worried about the ABI break. So before merging these changes, I have to check what was the ABI stability of Aravis during the 0.8 series, and see if this sort of breakage already happened. If not, I will probably switch to a new series. It may take a couple of weeks.

EmmanuelP:
It already happened in the 0.8 series a class definition has changed, so let's commit these changes.

Subclassing Aravis classes is not really supported. It should be made more clear by either making most of the class definition private (which would still allow subclassing, but will not permit to overwrite the virtual functions), or by simply moving the class definition in the private headers. That will not happen in the 0.8 series though.

Thanks a lot for your work on this issue. I had a look several time at it without success.

Cheers.

bpinsard:
Thanks a lot for the fix to both a you, just tested it and works perfectly. Will close the issue.

lostcontrol:
Thank you for your feedback. By the way, [Asyril SA](http://www.asyril.com), my employer, "sponsored" this fix by letting me work on it during my working time.

Issue #703, opened 2022-08-10
Handling arv_camera_[get|set]_[gain|black_level] with multiple taps


hackel6:
Some cameras have multiple sensor taps. As an example, Imperx GEV1620 has 2 taps. Each sensor tap has its own Black_Level and Analog Gain settings.
Using the current API for get and set will only return or change the value currently set via the "[Gain|BlackLevel]Selector" register which, in this case, can read either "AnalogTap1" or "AnalogTap2."
Thus, using the get/set API as it is right now only changes _half_ the image.
The other way to change both, aside from the *Selector registers, is to use each register directly: "AnalogOffset_Ch1" and "AnalogOffset_Ch2" respectively as an example (corresponds to BlackLevel).
Obviously, one possibility would be for the set() method to check for a corresponding *Selector register and set each value to the requested one at a time. However, the issue arises when a get() is performed and the taps (or channels) are _not_ all the same.

I'm more than willing to implement a solution, but I don't have a good idea how to handle it.

Another possibility would be for arv_camera_is_[gain|black_level]_available() to return false in the case where there are multiple taps, but that doesn't quite seem right either since it IS available and the proper use is to set each corresponding *Selector value to get/set the requested channel.

EmmanuelP:
I don't think it is a good idea to 'hide' the different gains in the current get/set API. And that would probably lead to strange behaviours withs camera that have overlapping gains (like All, Red, Blue, Green).

 I would add the following API instead :

* arv_camera_select_gain()
* arv_camera_dup_available_gains()

hackel6:
This makes sense to me and I may have some time soon to implement.
Just one question: i understand that just like the other *_dup_* calls already in the API, arv_camera_dup_available_gains() will return the available gain registers (probably as enumerations or strings). But just out of curiosity, what does "dup" stand for?

EmmanuelP:
`dup` is for duplicate. It is intended to indicate the returned data must be freed, as opposed to the `get` functions.

hackel6:
closed with #745

PR #745, opened 2022-12-15
Enhancement/703/multiple taps

hackel6:
#703


EmmanuelP:
Hi,

Sorry for the late reply.

I don't think we can select gain or black level using defined integer values. There is no guarantee a selector value is always mapped to the same string.

arv_camera_select_gain and arv_camera_select_black_level should use a string as the selector parameter.

src/arvcamera.c
```diff
@@ -1823,6 +1823,43 @@ arv_camera_get_gain_auto (ArvCamera *camera, GError **error)
...
+ *
+ * Configures Gain Selector feature.
+ *
+ * Since: 0.8.23
+ **/
+
+void
+arv_camera_select_gain (ArvCamera *camera, const char *gainSelector, GError **error) {
+	arv_camera_set_string (camera, "GainSelector", gainSelector, error);
+
```
EmmanuelP, line 1840: Opening bracket should on its own line. Closing bracket is missing.

src/arvcamera.c
```diff
@@ -1898,6 +1935,44 @@ arv_camera_get_black_level (ArvCamera *camera, GError **error)
...
+ * @blackLevelSelector: black level selection
+ * @error: a #GError placeholder, %NULL to ignore
+ *
+ * Configures Black Level Selector feature.
+ *
+ * Since: 0.8.23
+ **/
+
+void
+arv_camera_select_black_level (ArvCamera *camera, const char *blackLevelSelector, GError **error) {
```
EmmanuelP, line 1950: Opening bracket should on its own line.

EmmanuelP (changes requested):

src/arvcamera.c
hackel6: thanks, working too fast.

src/arvcamera.c
hackel6: thanks, working too fast.

hackel6:
@EmmanuelP, is this at least more of what you had in mind? raw strings? i guess it doesn't buy you all that much now since it just wraps to the arv_camera_set_string() method.

EmmanuelP (changes requested):
Sorry for the late review.

It looks good. Please update the Since statements in the function documentation.

EmmanuelP:
Thanks.

Issue #732, opened 2022-10-20
Spectral Instruments SI-1700 (RVT100) failing to take images with arv-camera-test


hackel6:
**Describe the bug**
We are trying to fold a new camera into our generic control architecture. We're finding that even with the simplest of test programs (arv-camera-tool), it seems to get the first frame and all others timeout. I have tried tweaking some of the GEV parameters listed [here](https://aravisproject.github.io/docs/aravis-0.8/ArvGvStream.html), but nothing seems to help with the timeouts. The same camera executes properly using Pleora eBUS SDK on Windows.

**Camera description:**
 - Spectral Instruments, Inc.-1700 (also known as RVT100)
 - Interface GigE

**Platform description:**
 - Aravis version: 0.8.19
 - OS: `Linux dvid-int-17 4.14.178-gentoo #17 SMP Thu May 27 17:04:21 PDT 2021 x86_64 Intel(R) Xeon(R) CPU E3-1220 V2 @ 3.10GHz GenuineIntel GNU/Linux`

**Additional context**
A short excerpt from the debug log:

```
[11:30:27.650] 🅳 genicam> [GcCommand::execute] AcquisitionStart (0x2)
[11:30:27.650] 🅳 stream-thread> [GvStream::find_frame_data] Start frame 1
[11:30:27.650] 🅳 misc> [PixelFormat::to_gst_caps_string] 0x01100005 -> video/x-raw, format=(string)GRAY16_LE
[11:30:27.650] 🅳 sp> packet_type  =       ok (0x0000)
                     content_type = data-leader (0x0001)
                     frame_id     =        1
                     packet_id    =        0
                     data_size    =       36
                     payload_type = image
                     pixel format = video/x-raw, format=(string)GRAY16_LE
                     width        = 200
                     height       = 200
                     x_offset     = 0
                     y_offset     = 0
[11:30:27.671] 🅳 sp> packet_type  =       ok (0x0000)
                     content_type = data-trailer (0x0002)
                     frame_id     =        1
                     packet_id    =       56
                     data_size    =        8
[11:30:27.671] 🅳 stream-thread> [GvStream::check_frame_completion] Completed frame 1
[11:30:27.671] 🅳 stream-thread> [GvStream::close_frame] Close frame 1
[11:30:27.671] 🅳 stream-thread> [GvStream::find_frame_data] Start frame 2
[...]

PR #746, opened 2022-12-16
Disable Leader packet timeout

hackel6:
#732 
https://aravis-project.discourse.group/t/external-triggering-of-spectral-instruments/548

Frames will no longer timeout after expiration of the frame_retention_timeout period if the ONLY packet received so far is a leader packet.


EmmanuelP:
Hi,

Thanks for the patch.

I think the test is incomplete, current_frame can point to a previous frame, for example if a resent packet is received.

hackel6:
does this look better? now i'm using last_frame_id to determine if it's the most recent frame received (which is really the only one we care about i think)

EmmanuelP:
Pushed to main. Thanks.

PR #747, opened 2022-12-19
added port legacy info for kowaOptronics SC130ET3

TomohisaAzegami:
SC130ET3 is one of our products. I'd like to my camera works on aravis. I hope that aravis project would accept this change.


EmmanuelP:
Thanks.

PR #769, opened 2023-02-27
Frame rate limit disabling also need vendor specific quirks

durnezj:
Hi,

I have the following fixes running successfully in my project as a patch and am now proposing to add these upstream so they can be included in a future release (and I can just update my buildroot's Aravis version).

This PR fixes the following:
- Vendor specific quirks to set the frame rate limit were only checked when frame rate limiting must be enabled.
- When wanting to disable frame rate limiting by calling arv_camera_set_frame_rate(cam,  **0**, NULL), the vendor specific quirks were not applied.
- This was observed on a FLIR and Basler camera, where the Basler works out of the box, but the the FLIR one needed a quirk to disable frame rate limiting.

I have proposed to fix this by extracting the code with the vendor specific quirks to a new method arv_camera_set_frame_rate_enable(cam, TRUE/FALSE, GError) that encapsulates this, and then call this in arv_camera_set_frame_rate().

(Now that I think of it, it might be good to provide a arv_camera_get_frame_rate_enable() as well?)

As it's my first time working with the GLib 'flow' and Aravis internals, please verify if everything is up to standards regarding code style, error handling,...

Kind regards,
Durnezj

### DIFF
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -1035,17 +1035,16 @@ arv_camera_set_frame_rate (ArvCamera *camera, double frame_rate, GError **error)
 	g_return_if_fail (ARV_IS_CAMERA (camera));
 
 	if (frame_rate <= 0.0) {
-		if (arv_camera_is_feature_available (camera, "AcquisitionFrameRateEnable", &local_error)) {
-			if (local_error == NULL)
-				arv_camera_set_boolean (camera, "AcquisitionFrameRateEnable", FALSE, error);
-			else
-				g_propagate_error (error, local_error);
+		arv_camera_set_frame_rate_enable(camera, FALSE, &local_error);
+		if (local_error != NULL)
+		{
+			g_propagate_error (error, local_error);
 		}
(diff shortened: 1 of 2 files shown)


durnezj:
I seem to have overlooked some leftover debug printf statements, will clean them up!

src/arvcamera.c
```diff
@@ -1271,6 +1259,66 @@ arv_camera_get_frame_rate_bounds (ArvCamera *camera, double *min, double *max, G
 	}
 }
 
+/*
+* arv_camera_set_frame_rate_enabled:
```
EmmanuelP, line 1263: The feature name in Genicam SFNC is AcquisitionFrameRateEnable. Please rename the function to arv_camera_set_frame_rate_enable.

EmmanuelP (changes requested):
Thanks for the patch.

It looks good, baring the function name.

durnezj:
Thanks for accepting the fix! :smile:

PR #777, opened 2023-03-20
Fix various warnings

feuerste:
This fixes a couple of warnings I got when locally building aravis. It also casts to the correct return types at various places.

Another noticeable change is that `arv_camera_dup_available_black_levels` was wrongly named in the corresponding `arvcamera.h`.

### DIFF
--- a/src/arvcamera.h
+++ b/src/arvcamera.h
@@ -141,8 +141,8 @@ ARV_API void		arv_camera_set_exposure_mode		(ArvCamera *camera, ArvExposureMode 
 
 ARV_API gboolean	arv_camera_is_gain_available		(ArvCamera *camera, GError **error);
 ARV_API gboolean	arv_camera_is_gain_auto_available	(ArvCamera *camera, GError **error);
-ARV_API void		arv_camera_select_gain				(ArvCamera *camera, const char *selector, GError **error);
-ARV_API const char **	arv_camera_dup_available_gains	(ArvCamera *camera, guint *n_selectors, GError **error);
+ARV_API void		arv_camera_select_gain			(ArvCamera *camera, const char *gainSelector, GError **error);
+ARV_API const char **	arv_camera_dup_available_gains	        (ArvCamera *camera, guint *n_selectors, GError **error);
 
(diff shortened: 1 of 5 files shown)


feuerste:
@EmmanuelP would you mind having a look? Thanks a lot!

src/arvcamera.h
```diff
@@ -141,8 +141,8 @@ ARV_API void		arv_camera_set_exposure_mode		(ArvCamera *camera, ArvExposureMode
 
 ARV_API gboolean	arv_camera_is_gain_available		(ArvCamera *camera, GError **error);
 ARV_API gboolean	arv_camera_is_gain_auto_available	(ArvCamera *camera, GError **error);
-ARV_API void		arv_camera_select_gain				(ArvCamera *camera, const char *selector, GError **error);
```
EmmanuelP, line 144: Please keep selector as the parameter name, and fix the c file

src/arvcamera.h
```diff
@@ -152,8 +152,8 @@ ARV_API ArvAuto		arv_camera_get_gain_auto		(ArvCamera *camera, GError **error);
 
 ARV_API gboolean	arv_camera_is_black_level_available	(ArvCamera *camera, GError **error);
 ARV_API gboolean	arv_camera_is_black_level_auto_available(ArvCamera *camera, GError **error);
-ARV_API void		arv_camera_select_black_level		(ArvCamera *camera, const char *selector, GError **error);
```
EmmanuelP, line 155: ditto

src/arvcamera.h
feuerste: Done.

src/arvcamera.h
feuerste: Done.

feuerste (commented):
Thanks for your review!

feuerste:
The problem in https://github.com/AravisProject/aravis/actions/runs/4529664926/jobs/7978355207?pr=777 seems to be unrelated to this PR...

EmmanuelP:
Thanks a lot.

PR #784, opened 2023-04-26
workaround not unique guid on some cams (DFM 37UX462-ML)

karelfv:
When plugin two and more such cams into the system, cams provide same guid (usually product id) and this confuses Aravis when accessing those so the only one seems to be available. The workaround is to concat serial number to the wrongly formed guid in such a case.

### DIFF
--- a/src/arvuvdevice.c
+++ b/src/arvuvdevice.c
@@ -37,6 +37,7 @@
 #include <arvstr.h>
 #include <arvzip.h>
 #include <arvmisc.h>
+#include <stdio.h>
 
 enum
 {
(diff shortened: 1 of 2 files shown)


EmmanuelP:
Hi @karelfv 

Sorry for the delay.

I'm not sure to understand the issue. If your cameras have the same GUID, you can still use the complete id to access them, like what is given by arv-tool-0.8. What is the output of arv-tools-0.8 when both devices are connected ?

EmmanuelP:
Hi @karelfv , could you please provide the output of `arv-tool-0.8` ?

karelfv:
Hi @EmmanuelP and sorry for delay. Here is requested info:
```
karel@rtems:~/vcs/aravis/build/src$ sudo ./arv-tool-0.8
The Imaging Source Europe GmbH-199E16220512-16220512 (USB3)
The Imaging Source Europe GmbH-199E16220523-16220523 (USB3)
karel@rtems:~/vcs/aravis/build/src$ 
```
This is on Debian 12 on amd64 and today's (2023-08-01) aravis tree.
So the thing seems to be working, am I right? Anyway, in the past I probably tested that on Alpine Linux or on older version of Debian. I may also tested that on arm64.

karelfv:
> It looks so. You can check the streaming part using `arv-camera-0.8`:
> 
> ```
> arv-camera-0.8 -n "The Imaging Source Europe GmbH-199E16220512-16220512" --duration 5
> arv-camera-0.8 -n "The Imaging Source Europe GmbH-199E16220523-16220523" --duration 5
> ```

This seems to be working too -- or well, it claims so. `arv-viewer-0.8` is still not working but this is known issue here.

PR #779, opened 2023-03-27
gv: bind incoming socket to device

EmmanuelP:
Instead of binding the socket to the interface address, use ADDR_ANY and set the BINDTODEVICE socket option.

- [ ] interface
- [ ] device
- [ ] stream

https://aravis-project.discourse.group/t/discover-camera-in-dhcp-mode/598/4

PR #819, opened 2023-09-08
Discover dhcp camera outside subnet

hackel6:
When possible (running as root) uses SO_BINDTODEVICE to discover cameras outside the subnet.
I did not implement for Stream because the idea is that with this change you can now use arv-tool-0.8 (with root privelages) to re-address the camera and bring it back into communication for proper use.
I'm not sure if there's a better way that does not require root for SO_BINDTODEVICE.
https://aravis-project.discourse.group/t/discover-camera-in-dhcp-mode/598/4
https://github.com/AravisProject/aravis/pull/779

draft until i can test next week

### DIFF
--- a/src/arvgvdevice.c
+++ b/src/arvgvdevice.c
@@ -1966,7 +1966,7 @@ arv_gv_device_constructed (GObject *object)
 	char *address_string;
 	guint32 capabilities;
 	guint32 device_mode;
-	GSocketAddress *any_socket_address;
+	GSocketAddress *socket_address;
 	GInetAddress *any_address;
 	int socket_fd;
 	struct ifaddrs *addrs, *iap;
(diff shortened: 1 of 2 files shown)


hackel6:
You mentioned in #779 that SO_BINDTODEVICE is Linux only. I suppose I could surround with #ifdef statements; unless you think there's a better way. If i have time i can experiment with other broadcast options.

hackel6:
I was unable to get the response broadcast packets to be picked up by the receiving socket without SO_BINDTODEVICE. it's possible there is a linux kernel setting which will allow them to get through, but i was unable to figure it out. Reverse Path Filtering is turned off by default on the systems i'm testing with, so I'm kind of out of ideas and I suppose have to use the approach from this PR and run with root in order to use SO_BINDTODEVICE.

hackel6:
it looks like this MR is now overcome by events. I think it can be closed. i have tested camera discovery via LLA by running arv-tool as root under 0.8.35 and it appears to be working without this change.

PR #822, opened 2023-09-23
Add AutomationTechnology C6 to legacy exceptions; add ARV_QUIRK_LEGACY_ENDIANNESS env var

eudoxos:
Please have a look what it feels like. I don't feel strong about the env var workaround (and the name could be also different); for quick testing, it could work better than repeated instruction "add to the exception list and recompile".

### DIFF
--- a/src/arvgcport.c
+++ b/src/arvgcport.c
@@ -146,6 +146,8 @@ _use_legacy_endianness_mechanism (ArvGcPort *port, guint64 length)
 
 		if (arv_gc_register_description_node_compare_schema_version (register_description, 1, 1, 0) < 0) {
 			port->priv->has_legacy_infos = TRUE;
+		} else if (g_getenv("ARV_QUIRK_LEGACY_ENDIANNESS")) {
+			port->priv->has_legacy_infos = TRUE;
 		} else {
 			for (i = 0; i < G_N_ELEMENTS (arv_gc_port_legacy_infos); i++) {
 				if (g_pattern_match_simple(arv_gc_port_legacy_infos[i].vendor_selection, vendor_name) == TRUE &&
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Thanks. I will commit the new entry.

I don't really want to make it easy to enable the legacy workaround, as I'd like to be kept informed about the devices requiring the workaround. Once a user has patched aravis, there is more chance I will get a merge request.

eudoxos:
For me it is okay this way. I understand your reasoning as well. I just think that there would be more people without resources to recompile (time, motivation, skills, ...) who will just leave Aravis if they don't have a quick way to override this.

EmmanuelP:
Pushed the first commit to main.

Thanks.

eudoxos:
I would have a compromise solution. Use the environment variable, but then display an annoying warning like this:
```
!!! WARNING:
!!! ARV_QUIRK_LEGACY_ENDIANNESS is being used.
!!! This is meant for debugging-only.
!!! Please submit your camera details to
!!!    https://github.com/AravisProject/aravis/issue/new?...
!!! It will help other Aravis users and this message will disappear.
!!! Thank you.
```
Don't forget that many users these days have Aravis pre-packaged (in my case both msys2 and debian), and recompilation is really not easy for everybody.

EmmanuelP:
Event better, I think we should be able to detect when the quirk is required, by reading a 4 byte register with the 2 methods (register and memory read). I have to experiment a bit with this idea.

eudoxos:
I mentioned that in one of my (chaotic) replies in the discussion [here](https://aravis-project.discourse.group/t/automationtechnology-c6/664/4): would the `GevDeviceModeIsBigEndian` node be of any use? I skimmed genicam data of some of the cameras on the exception list (here in issue reports) and a few of them actually had that node.

A possible heuristics would be to read a 2-byte range-restricted node and see which endianness yields a value within range (such as `Width` or `Height`).

Issue #823, opened 2023-09-28
Aravis won't correctly connect to cameras over a VPN


jsarrett:
**Describe the bug**
arv_enumerate_network_interfaces ignores Point to Point IP links when discovering interfaces.  This causes packets destined for remote networks across that link to have the wrong source address.

**To Reproduce**
1. connect to a network over a Point to point link, e.g. openconnect.
2. run arv-tool-0.8 -a <Address reachable on that link>

**Expected behavior**
arv-tool-0.8 connects and prints out the remote Address

**Camera description:**
 - Manufacturer Any
 - Model Any
 - Interface  GigEVision

**Platform description:**
 - Aravis version 0.8 (main)
 - OS: Tested on Ubuntu 23.04
 - Hardware x86_64
This is a system behavior problem which will affect pretty much all UNIX like platforms

**Additional context**
Aravis has a nasty habit of setting source IPs from it's interface discovery, which doesn't necessarily respect the system IP routing table.  When Aravis discounts a perfectly usable IP link, but you specify an address to connect to, it will pick a source IP incorrectly, and send a packet that potentially can't be returned.

jsarrett:
PR #824 addresses this.

PR #824, opened 2023-09-28
Don't ignore POINTOPOINT interfaces on UNIX family OSes

jsarrett:
This patch allows Aravis to connect to cameras over POINTOPOINT IP links.  Broadcasts are not expected to work, but unicast does.  This allows connecting to cameras across a VPN, or PPPoE network.  This addresses issue #823


EmmanuelP:
Merged. Thanks.

PR #831, opened 2023-10-21
Add a GenTL consumer interface

xiaoqiangwang:
It makes aravis a GenTL consumer and work with GenICam cameras with non USB/Ethernet interfaces .

One breaking change is that GenTL requires its own start/stop acquisition calls, which have to be synched with ArvCamera object, i.e.
```
# prepare parameters
...
# start
arv_stream_start_acquisition
arv_camera_start_acquisition

# stop
arv_camera_stop_acquisition
arv_stream_stop_acquisition
```

I have tested the implementation with a Basler acA1300-200uc USB3 camera with Pylon SDK 7.3 on a Debian 12 x86_64 system. Both arv-tool and arv-viewer work normally.  https://github.com/areaDetector/ADAravis can also work with this new implementation.


xiaoqiangwang:
One more thing to note is arv-viewer crashes when loading *ProducerU3V.cti* or *ProducerGEV.cti*
```
(gdb) bt
#0  0x00007ffff6cfcd3c in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff6cadf32 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff6c98472 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007ffff5aba3ab in ?? () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#4  0x00007ffff5acdff1 in _Unwind_RaiseException () from /lib/x86_64-linux-gnu/libgcc_s.so.1
BFD: reopening /tmp/shared.uxJxTn: No such file or directory
warning: Can't read data for section '.eh_frame' in file '/tmp/shared.uxJxTn'
#5  0x00007fffd557ea7c in ?? () from /tmp/shared.uxJxTn
#6  0x00007fffffffbea0 in ?? ()
#7  0x00007fffffffbee0 in ?? ()
#8  0x0000555555980780 in ?? ()
#9  0x00007fffd54ad2b0 in ?? () from /tmp/shared.uxJxTn
#10 0x00007fffd56547f0 in ?? () from /tmp/shared.uxJxTn
#11 0x0000555555bb7868 in ?? ()
#12 0x0000000000000000 in ?? ()
```

arv-tool and ADVimba do not have this problem. I did not dig into the exact cause yet.

EmmanuelP:
Hi @xiaoqiangwang ,

Thanks a lot. I just ran a quick test that worked fine using Basler U3VProvider. Nice work !

> One breaking change is that GenTL requires its own start/stop acquisition calls, which have to be synched with ArvCamera object, i.e.

Question about that: is the execution of the Genicam feature AcquisitionStart supposed to start the acquisition on all the device datastream ? If so, we can keep a track off all created streams, and call stream_start_acquisition internally. You may use g_object_weak_ref from the ArvDevice to keep this list.

Also, it would be nice to add a arv_stream_create_and_push_buffers that will create the buffers knowing the datastream instance, and possibly help to avoid the memory copy. We will also need a arv_buffer_new_with_user_data for that purpose.

EmmanuelP:
> > If so, we can keep a track off all created streams, and call stream_start_acquisition internally. You may use g_object_weak_ref from the ArvDevice to keep this list.
> 
> If I create two news ArvDevice methods, device_start/stop_acquisition, these two relay camera_start/stop_acquisition calls to underlying stream_start/stop_acquisition. Would that be Ok?

Yes, seems fine.

EmmanuelP:
> > Also, it would be nice to add a arv_stream_create_and_push_buffers that will create the buffers knowing the datastream instance, and possibly help to avoid the memory copy. We will also need an arv_buffer_new_with_user_data for that purpose.
> 
> I understand this need but have no confidence in designing this API.

No problem. Once your work is merged, we can iterate and fix these sort of issues.

xiaoqiangwang:
The change to start/stop_acquisition has been made. Since the relay is internalised, user API is unaffected. arvviewer.c and arvtest.c are unmodified and the tests pass. Thanks for the suggestion.

I have briefly tested the idea of adding buffers directly to GenTL producers, https://github.com/xiaoqiangwang/aravis/tree/gentl_buffer. But it turns out that Basler GenTL producers do not support adding buffers during acquisition. And GenTL standard does not make it mandatory.

For applications with a static set of buffers, that may work. But in [this case](https://github.com/areaDetector/ADAravis/blob/master/aravisApp/src/ADAravis.cpp#L713), it constantly passes on filled buffers and created and queue new ones.

EmmanuelP:
Do you still have pending work on this pull request ? Any known crash or bug ?

xiaoqiangwang:
I am testing the event notification. The reading of event data is there. For the user API, I imagine a signal "device-event" on the device level. 

We can also make the event notification another PR.

xiaoqiangwang:
I have pushed the event notification part. arv-test has a new *Event* option for testing.

One problem with the event data nodes is that they stopped reading after the first event update. Is there a way to invalidate the node or force a new read?

xiaoqiangwang:
For completeness I did a test on Windows.

There are two warnings. In the first the signals are not recognized.
```
> arv-camera-test-0.8.exe -n "Basler acA1300-200uc (23524298)"
Looking for camera 'Basler acA1300-200uc (23524298)'
vendor name            = Basler
model name             = acA1300-200uc
device serial number   = 23524298
image width            = 1280
image height           = 1024
exposure               = 5000 ┬╡s
gain                   = 0 dB
payload                = 1310720 bytes

(arv-camera-test-0.8.exe:5644): GLib-GObject-WARNING **: 13:11:55.246: ../src/gobject/gsignal.c:2620: signal 'new-buffer' is invalid for instance '0000018051BCC0A0' of type '(NULL)'

(arv-camera-test-0.8.exe:5644): GLib-GObject-WARNING **: 13:11:55.246: ../src/gobject/gsignal.c:2620: signal 'control-lost' is invalid for instance '0000018051154730' of type '(NULL)'
  0 frame/s  -       0 MiB/s
  0 frame/s  -       0 MiB/s
  0 frame/s  -       0 MiB/s
  0 frame/s  -       0 MiB/s
  0 frame/s  -       0 MiB/s
  0 frame/s  -       0 MiB/s
  0 frame/s  -       0 MiB/s
  0 frame/s  -       0 MiB/s
n_completed_buffers    = 41
n_failures             = 0
n_underruns            = 0
n_transferred_bytes    = 53739520

(arv-camera-test-0.8.exe:5644): GLib-CRITICAL **: 13:12:03.370: g_datalist_id_set_data_full: assertion 'key_id > 0' failed

(arv-camera-test-0.8.exe:5644): GLib-CRITICAL **: 13:12:03.370: g_datalist_id_set_data_full: assertion 'key_id > 0' failed
```

The conan packages used are
```
> conan info -if build . --only None
Generated graphinfo
conanfile.txt
bzip2/1.0.8
glib/2.74.1
gstreamer/1.22.3
libffi/3.4.4
libgettext/0.21
libiconv/1.17
libusb/1.0.26
libxml2/2.10.3
meson/0.59.2
ninja/1.10.2
pcre2/10.42
pkgconf/1.9.3
zlib/1.2.13
```

xiaoqiangwang:
@EmmanuelP just to be clear that this PR is ready for review. It passes the arv-test program and runs stable with ADAravis.

EmmanuelP:
Ok. I will try to merge your work next week. Thanks again.

EmmanuelP:
Hi @xiaoqiangwang 

I have commited most of your changes in main. Events are disabled for now (the code is in, but commented out), I have still to think about the API. The changes in arv-test are not commited, the addition of the event testing in the transfer test code makes thing a bit too complicated to my taste.

Thanks again, nice piece of work !

PR #833, opened 2023-10-31
added get/set register node value

Fixstars-momoko:
I have implemented an API to set/get the value of the "Register-Type" node in order for the Aravis library to handle cameras in the GENDC data format in the future. 
Currently, when using the arv-tool to display the GenDC descriptor, it appears as follows:

```
$ ./arv-tool-0.8.exe -n  <name of camera> control GenDCDescriptor
Length of GenDCDescriptor = 1280
Content of GenDCDescriptor =
        0x47 0x4e 0x44 0x43 0x01 0x00 0x00 0x00
...
```

Note that
* gendc is supported as U3V camera's payload type from U3V v1.2
* `GNDC` (`0x47 0x4e 0x44 0x43`) is the signature of GenDCDescriptor of the beginning of U3V payload
* I only have U3V camera whose data format is GenDC; I have never tested on GigE.

Reference: https://www.emva.org/wp-content/uploads/GenICam_GenDC_v1_1.pdf

### DIFF
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -2863,9 +2863,9 @@ arv_camera_get_register (ArvCamera *camera, const char *feature,guint64 length, 
  * arv_camera_get_register_length:
  * @camera: a #ArvCamera
  * @feature: feature name
- * @error: a #GError placeholder, %NULL to ignore
+ * @error: a #GError placeholder
  *
- * Returns: the length of register value.
+ * Returns: the length of register value, 0 if not available
  *
  * Since:
  */
@@ -2875,7 +2875,7 @@ arv_camera_get_register_length (ArvCamera *camera, const char *feature, GError *
 {
 	ArvCameraPrivate *priv = arv_camera_get_instance_private (camera);
 
-	g_return_val_if_fail (ARV_IS_CAMERA (camera), 0.0);
+	g_return_val_if_fail (ARV_IS_CAMERA (camera), 0);
 
 	return arv_device_get_register_feature_length (priv->device, feature, error);
 }
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -2854,7 +2854,7 @@ arv_camera_get_register (ArvCamera *camera, const char *feature,guint64 length, 
 {
 	ArvCameraPrivate *priv = arv_camera_get_instance_private (camera);
 
-	g_return_val_if_fail (ARV_IS_CAMERA (camera));
+	g_return_if_fail (ARV_IS_CAMERA (camera));
 
 	arv_device_get_register_feature_value (priv->device, feature, length, value, error);
 }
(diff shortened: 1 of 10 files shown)


src/arvtool.c
```diff
@@ -364,7 +370,28 @@ arv_tool_control (int argc, char **argv, ArvDevice *device)
...
+                                                buffer = (unsigned char*) malloc(length);
+                                                arv_gc_register_get (ARV_GC_REGISTER (feature), (void*)buffer, length, &error);    
+
+                                                if (error == NULL){
+                                                        printf("Content of %s =", tokens[0]);
+                                                        for( int i = 0; i < length; i++){
+                                                                if ( i%8 == 0){
+                                                                        printf("\n\t");
+                                                                }
+                                                                printf("0x%02x ", *((buffer)+i));
```
EmmanuelP, line 389: Use arv_g_string_append_hex_dump here.

Please format the output as:
```
GenDCDescriptor = 128 bytes@0x0900
01e0 c8 b7 89 b0 45 fa 3d 9d 8c e9 a7 33 46 85 1f 2c  ....E.=....3F..,
...
```

src/arvdevice.c
```diff
@@ -798,6 +799,72 @@ arv_device_get_float_feature_increment (ArvDevice *device, const char *feature,
...
+arv_device_get_register_feature_length (ArvDevice *device, const char *feature, GError **error)
+{
+	ArvGcNode *node;
+	node = _get_feature (device, ARV_TYPE_GC_REGISTER, feature, error);
+	if (node != NULL)
+		return arv_gc_register_get_length(ARV_GC_REGISTER (node), error);
+
+	return 0;
+}
+
```
EmmanuelP, line 867: Same remark here about the API simplification. In arvDevice, the get_register function should take care to check the buffer length value is lower than the maximum size (2^16), and return an error otherwise.

EmmanuelP (changes requested):
Thanks. Some minor things to fix before it can go in.

src/arvcamera.c
```diff
@@ -2816,6 +2817,69 @@ arv_camera_get_float_increment (ArvCamera *camera, const char *feature, GError *
...
+guint64
+arv_camera_get_register_length (ArvCamera *camera, const char *feature, GError **error)
+{
+	ArvCameraPrivate *priv = arv_camera_get_instance_private (camera);
+
+	g_return_val_if_fail (ARV_IS_CAMERA (camera), 0);
+
+	return arv_device_get_register_feature_length (priv->device, feature, error);
+}
+
```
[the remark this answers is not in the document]
EmmanuelP: And rename get_register to dup_register, which makes clear the data must be freed after use.

EmmanuelP:
I have made the changes and merged. Thanks.

PR #834, opened 2023-11-02
Gendc/u3v single image component

Fixstars-momoko:
USB 3 Vision Camera started to support GenDC (https://www.emva.org/wp-content/uploads/GenICam_GenDC_v1_1.pdf) as one of the payload type.

GeDC is generic data container, which consists of Descriptor and Data, can store multiple/multimedia in a single container for each Data component, and its mapping and structure design is noted in Descriptor.

## Constraints

In this PR, I have not covered some devices and features:

### U3V

I confirmed that arv-viewer successfully displays the image in the first component of GenDC contaner from U3V camera.

I have never try this on GigE since I do not have any GigE  camera whose data format is GenDC.

### Only first component data is supported 

Since Aravis currently does not support multipart sensor data for U3V, I just pick the data if it is 

1. in the first available (valid) component 
2. `TypeId ` (one of SFNC’s `ComponentIdValue `feature) is image = Intensity(1)  reference: https://www.emva.org/wp-content/uploads/GenICam_SFNC_v2_7.pdf

I confirmed that the image data can be retrieved by `arv_buffer_get_part_data` and the whole GenDC container (Descriptor +Data)  `arv_buffer_get_data`.

### U3V Payload Type

This PR only supports GenDC Container ( Payload Type 0x1001 on U3V) but not GenDC Component Data  ( Payload Type 0x1002 on U3V).

Currently U3V (v1.2) supports 5 payload types: Image, Image Extended Chunk, Chunk, GenDC Container, and GenDC Component Data.

---

@EmmanuelP What do you think about this change? I would be happy if Aravis supports GenDC devices, but I also know that it is very hard to find GenDC devices to test this.


EmmanuelP:
> @EmmanuelP What do you think about this change? I would be happy if Aravis supports GenDC devices, but I also know that it is very hard to find GenDC devices to test this.

That can definitely go into aravis.  Is really the multipart API a good fit to access GenDC data, or do we need a dedicated set of functions ?

EmmanuelP:
Ok, let's merge your work. If the arvbuffer part API  is not the right one, we can add a specific GenDC API later without breaking anything (I hope).

EmmanuelP:
Thanks.

PR #848, opened 2023-11-23
producer: MSVC compilation fixes

EmmanuelP:


eudoxos:
Please integrate this commit, it fixes w32 https://github.com/eudoxos/aravis/commit/e50560e3d69706cdf6db331a445286bebc339768 .

Below, FWIW, commands for cross-compilation (though the CI does it natively), for very quick testing.


# Crossroad

Install like this (`pip3 install --user git+git://...` would not work for some reason):
```
git clone git://git.tuxfamily.org/gitroot/crossroad/crossroad.git
cd crossroad; pip3 install -r requirements.txt
python3 setup.py install --user
```

Enter crossroad (`w32` will use `mingw-w64-i686` arch):
```
crossroad w32 aravis
```

Install dependencies:
```
crossroad install glib2 libxml2
```

# Build Aravis

Navigate to Aravis sources and configure the build:
```
crossroad meson . build-w32 -Dbuildtype=debug -Ddocumentation=disabled -Dgst-plugin=disabled -Dintrospection=disabled -Dusb=disabled -Dviewer=disabled -Dgentl-producer=true
```
Build:
```
ninja -C build-w32
```

eudoxos:
And then, there is a bunch of warnings about printf formating. What is the status of [inttypes.h](https://en.wikibooks.org/wiki/C_Programming/inttypes.h)? IIRC we discussed it at some point already, concluding that it might not be supported everywhere. Has that change? Then most of the warnings could be fixed.

EmmanuelP:
Thanks

PR #850, opened 2023-11-27
Rename typeid variable name.

feuerste:
`typeid` is also an operator, see https://en.cppreference.com/w/cpp/language/typeid. This renaming makes sure they don't get mixed up.


EmmanuelP:
Thanks.

PR #851, opened 2023-11-27
Add GenTL namespace when using respective code.

feuerste:
When compiling code with a C++ compiler the `GenTL_v1_5.h` header declares a couple of things inside an additional `GenTL` namespace.
If we compile for C++, we also need to use this namespace in order to compile successfully.

### DIFF
--- a/src/arvgentldeviceprivate.h
+++ b/src/arvgentldeviceprivate.h
@@ -31,11 +31,17 @@
 #include <arvgentlsystemprivate.h>
 
 G_BEGIN_DECLS
+#ifdef G_CXX_STD_VERSION
+namespace GenTL {
+#endif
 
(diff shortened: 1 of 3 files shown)


EmmanuelP:
Thanks.

EmmanuelP:
I may have pulled the trigger too fast. How comes this patch is needed ? Private headers are not installed, aravis is compiled using a C compiler.

feuerste:
> I may have pulled the trigger too fast. How comes this patch is needed ? Private headers are not installed, aravis is compiled using a C compiler.

My bad and apologies, we have a bazel build setup for aravis independent from meson and accidentally created `arvenumtypes[private].cc` instead of `arvenumtypes[private].c`, triggering the c++ compiler... I will create a revert PR.

PR #857, opened 2024-01-10
install arvnetwork.h header

hesmar:
The header `arv.h` includes `arvnetwork.h` but this file gets not installed. This PR adds `arvnetwork.h` to `library_headers` to fix the issue.


EmmanuelP:
Thanks.

PR #860, opened 2024-01-27
add NIT NSC1601GIGE to Genicam 1.0 legacy exception list

jwahlstrand:
This allows using a NIT WiDy SenS 640 camera with aravis.


EmmanuelP:
Thanks.

PR #876, opened 2024-03-11
Fix passing of AcquisitionFrameRateEnable boolean flag.

feuerste:
Previously for unknown and a couple of other cameras the actual flag value was ignored and always set to true.


EmmanuelP:
Thanks

PR #877, opened 2024-03-11
Reuse arv_camera_set_frame_rate_enable method when setting trigger. 

feuerste:
This way we can support a wider range of cameras with e.g. different naming schemes for the `AcquisitionFrameRateEnable` feature.


EmmanuelP:
Thanks.

PR #880, opened 2024-03-11
xml: make error struct constant

russelltg:
As of https://gitlab.gnome.org/GNOME/libxml2/-/commit/61034116d0a3c8b295c6137956adc3ae55720711, the error struct in the callback is const

I could also add a cast to `xmlStructuredErrorFunc` on the use sites so it compiles against both new and old versions of libxml without warnings/errors.


EmmanuelP:
Thanks.

PR #881, opened 2024-03-12
Fix compilation with libxml older than 2.12.

feuerste:
#880 introduced compatibility with libxml 2.12, but also caused compilation errors on systems with only 2.11.
This fixes compilation for systems with older libxml versions.


EmmanuelP:
Thanks

PR #886, opened 2024-03-15
Also test `ArvGevSCPSFireTestPacket` for auto packet size.

feuerste:
Some cameras, e.g. the Allied Vision Alvium G1, don't have the `GevSCPSFireTestPacket` feature, but only the `ArvGevSCPSFireTestPacket`, so we now test both.

### DIFF
--- a/src/arvgvdevice.c
+++ b/src/arvgvdevice.c
@@ -647,6 +647,7 @@ auto_packet_size (ArvGvDevice *gv_device, gboolean exit_early, GError **error)
 	ArvGvDevicePrivate *priv = arv_gv_device_get_instance_private (gv_device);
 	ArvDevice *device = ARV_DEVICE (gv_device);
         ArvGcNode *node;
+	gboolean has_test_packet = 0;
 	GSocket *socket;
 	GInetAddress *interface_address;
 	GSocketAddress *interface_socket_address;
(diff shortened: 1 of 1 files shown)


feuerste:
@EmmanuelP I just saw a comment in e30bdaf064a5e12434f448b3cc61e707260d91bd about  `GevSCPSFireTestPacket`. I guess my PR may not make sense, as the `Arv` prefixed version is only internal to aravis, right? However, the `test_packet_check` function also uses the prefixed version. With the changes in this PR the Allied Vision Alvium G1 is able to do automatic packet size adjustments, without it it isn't.
Would you mind to explain what exactly happens with the prefix? Thanks a lot!

feuerste:
> @EmmanuelP I just saw a comment in [e30bdaf](https://github.com/AravisProject/aravis/commit/e30bdaf064a5e12434f448b3cc61e707260d91bd) about `GevSCPSFireTestPacket`. I guess my PR does not make sense, as the `Arv` prefixed version `ArvGevSCPSFireTestPacket` is only internal to aravis and always available, right? However, the `test_packet_check` function also uses the prefixed version. With the changes in this PR (basically resulting in the feature check to be always true) the Allied Vision Alvium G1 is able to do automatic packet size adjustments, without it it isn't. Would you mind to explain what exactly happens with the prefix? Thanks a lot!

Thinking more about this, would a viable alternative to checking whether `GevSCPSFireTestPacket` is available be that we remove the check and add a timeout, so the `auto_packet_size` function doesn't take more than a second or couple of seconds?

feuerste:
Closing this in favor of #888.

PR #888, opened 2024-03-21
Add timeout for auto packet size determination.

feuerste:
Previously, we relied on the availability of the `GevSCPSFireTestPacket` feature. By adding a timeout we can now also support cameras without this feature.

This is a follow-up to #886 and, as mentioned in e30bdaf064a5e12434f448b3cc61e707260d91bd,  should also be able to deal with cameras such as the DMK 23G618 which takes 300ms to complete a single fire test packet command using `ArvGevSCPSFireTestPacket`.

### DIFF
--- a/src/arvgvdevice.c
+++ b/src/arvgvdevice.c
@@ -646,7 +646,6 @@ auto_packet_size (ArvGvDevice *gv_device, gboolean exit_early, GError **error)
 {
 	ArvGvDevicePrivate *priv = arv_gv_device_get_instance_private (gv_device);
 	ArvDevice *device = ARV_DEVICE (gv_device);
-        ArvGcNode *node;
 	GSocket *socket;
 	GInetAddress *interface_address;
 	GSocketAddress *interface_socket_address;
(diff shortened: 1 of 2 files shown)


src/arvgvdevice.c
```diff
@@ -726,6 +723,11 @@ auto_packet_size (ArvGvDevice *gv_device, gboolean exit_early, GError **error)
 		guint current_size = packet_size;
 
 		do {
+			if (g_timer_elapsed (timer, NULL) > 1.0) {
```
feuerste, line 726: We could also make this timeout configurable.
I was just not sure:
1) Whether we want this at all.
2) Whether we want to change the APIs of `arv_gv_device_auto_packet_size`/`arv_camera_gv_auto_packet_size` for this or
3) Want to introduce new methods, e.g. `arv_gv_device_auto_packet_size_timeout`/`arv_camera_gv_auto_packet_size_timeout`

feuerste:
@EmmanuelP Do you have any thoughts on this PR? Looking forward to hearing from you.

EmmanuelP:
I don't know if the test packet is a mandatory feature of GigEVision. If I recall correctly, I have added the `GevSCPSFireTestPacket` presence test because one of the camera I own did not send test packets.

That means devices not supporting test packets will have an additional 1 second startup delay.

I would prefer the addition of an `options` parameter to `arv_gv_device_auto_packet_size`. Aravis will have API breaks in the transition to 0.10 stable series, so it is the right time to break this API.

`ARV_GV_DEVICE_AUTO_PACKET_SIZE_OPTION_FORCE` will allow to bypass the feature presence test.

feuerste:
Thanks Emmanuel for your answer and thoughts. If there are cameras out in the wild which don't support test packets at all then I would rather like to keep it the way it is now and close this PR. Thanks again!

PR #889, opened 2024-03-21
device: use const void* in `arv_device_write_memory`

freeman94:
The purpose of this PR is to be able to generate a corresponding [FFI binding in Rust](https://docs.rs/aravis/latest/aravis/) without needing `mut` on the buffer to be written.


EmmanuelP:
Thanks

PR #895, opened 2024-04-08
Fix finalization of gv device.

feuerste:
Without the fix after camera creation, `DeviceReset`, device list update, and camera recreation a segmentation fault happened because of a null pointer when destructing the original camera.


EmmanuelP:
Thanks

PR #896, opened 2024-04-09
Fix `arv_buffer_get_chunk_data` for extended chunk data.

feuerste:
Before the fix with a FLIR Blackfly S BFS-PGE-23S3C in chunk mode I received lots of `** (process:1179494): CRITICAL **: 08:18:03.408: arv_buffer_get_chunk_data: assertion 'arv_buffer_has_chunks (buffer)' failed` errors, as this camera only supports `ARV_BUFFER_PAYLOAD_TYPE_EXTENDED_CHUNK_DATA`.

### DIFF
--- a/src/arvgvspprivate.h
+++ b/src/arvgvspprivate.h
@@ -377,7 +377,8 @@ arv_gvsp_leader_packet_get_buffer_payload_type (const ArvGvspPacket *packet, gbo
 
                 if (has_chunks != NULL)
                         *has_chunks = ((payload_type & 0x4000) != 0 ||
-                                       (payload_type == 0x0004));
+                                       (payload_type == ARV_BUFFER_PAYLOAD_TYPE_CHUNK_DATA) ||
+                                       (payload_type == ARV_BUFFER_PAYLOAD_TYPE_EXTENDED_CHUNK_DATA));
 
                 return (ArvBufferPayloadType) (payload_type & 0x3fff);
         }


feuerste:
@EmmanuelP We may also need to adapt https://github.com/AravisProject/aravis/blob/853c5d101a3a94b18648bcfcea7e5abd5a2f5791/src/arvuvspprivate.h#L181 to the same logic for `has_chunks`, but in https://github.com/AravisProject/aravis/blob/21da082c64b94360a73797c2b26cafc1b9b992d1/src/arvuvstream.c I don't find any occurrences of `ARV_BUFFER_PAYLOAD_TYPE_CHUNK_DATA`, so more work may be needed there. What do you think?

EmmanuelP:
> @EmmanuelP We may also need to adapt
> 
> https://github.com/AravisProject/aravis/blob/853c5d101a3a94b18648bcfcea7e5abd5a2f5791/src/arvuvspprivate.h#L181

Here's what is in wrieshark u3v dissector:

```
  413 #define U3V_STREAM_PAYLOAD_IMAGE            0x0001
  414 #define U3V_STREAM_PAYLOAD_IMAGE_EXT_CHUNK  0x4001
  415 #define U3V_STREAM_PAYLOAD_CHUNK            0x4000
```

So the test for the chunk bit is enough in this case.

EmmanuelP:
Thanks for the patch.

Issue #894, opened 2024-04-02
Lots of missing packets talking to a GigE camera: debugging suggestions?


dkogan:
Hi. I'm seeing a consistent issue where my camera cannot receive even one successful image due to missing gige packets. It's similar to #434. I can do my own debugging, but I'm hoping you can suggest some better ways to do that. Observations:

- High-resolution GigE camera (Emergent HR-20000C; 20MP)
- Linux. Fast machine, but very old kernel (3.10)
- No missing packets when running the vendor's tools (they work)
- No missing packets in tcpdump logs
- Running as root enables the packet-socket method, and that fixes the problem. But the standard socket method is what is expected to be used most of the time, and that's supposed to work too; right?
- MTU is already 9000, and increasing the socket buffer size and retention time doesn't appear to make a difference
- Poking at the code, it looks like the packets are lost in the OS somewhere, not in aravis

So given all that, any suggestions? In your experience, are the packets being lost in the OS? If so, it should be possible to use perf, or something, to see conclusively where the packet loss is occurring. What do you do, usually?

Thank you very much for aravis. It's truly incredible. I'm now porting the 3rd set of my cameras to use it, and more or less, it's working perfectly.

dkogan:
I poked at this a bit more. In short, I wasn't setting the socket buffer size properly, and fixing that made it work.

Notes:

The socket buffer size can be set like this:

```
g_object_set (*stream,
              "socket-buffer",      ARV_GV_STREAM_SOCKET_BUFFER_FIXED,
              "socket-buffer-size", 20000000,
              NULL);
```

This sets the `SO_RCVBUF` setting on the socket. Described like this in
socket(7):

```
SO_RCVBUF
    Sets or gets the maximum socket receive buffer in bytes. The
    kernel doubles this value (to allow space for bookkeeping
    overhead) when it is set using setsockopt(2), and this doubled
    value is returned by getsockopt(2). The default value is set by
    the /proc/sys/net/core/rmem_default file, and the maximum allowed
    value is set by the /proc/sys/net/core/rmem_max file. The minimum
    (doubled) value for this option is 256.
```

The setting ends up in `sk_rcvbuf` in the kernel:

- https://lxr.linux.no/#linux+v6.7.1/net/core/sock.c#L1250
- https://lxr.linux.no/#linux+v6.7.1/net/core/sock.c#L960

Which is used in multiple places in https://lxr.linux.no/#linux+v6.7.1/net/ipv4/udp.c

[...]

PR #898, opened 2024-04-18
Better documentation, error messages, defaults regarding the gige socket-buffer-size

dkogan:
These are the result of me encountering and debugging packet-dropping issues: https://github.com/AravisProject/aravis/issues/894

Primarily these patches add more error reporting and documentation. With the extended error messages I think it makes sense to change the default socket-buffer policy to `ARV_GV_STREAM_SOCKET_BUFFER_AUTO`, which this patch series does.

There's one more thing I would change also, but want to get your thoughts about all this first:

If setsockopt() returns success, but doesn't actually give us the requested buffer, we currently:

- Report a `arv_warning_interface()`
- `arv_socket_set_recv_buffer_size()` returns FALSE
But the caller of `arv_socket_set_recv_buffer_size()` doesn't do anything with this FALSE, and things continue as normal. I would make this a hard error, and I'd change the `arv_warning_interface()` to something that would be seen even without any `ARV_DEBUG` flags.

Thanks much

### DIFF
--- a/src/arvgvstream.c
+++ b/src/arvgvstream.c
@@ -1831,7 +1831,7 @@ arv_gv_stream_class_init (ArvGvStreamClass *gv_stream_class)
 		g_param_spec_enum ("socket-buffer", "Socket buffer",
 				   "Socket buffer behaviour",
 				   ARV_TYPE_GV_STREAM_SOCKET_BUFFER,
-				   ARV_GV_STREAM_SOCKET_BUFFER_FIXED,
+				   ARV_GV_STREAM_SOCKET_BUFFER_AUTO,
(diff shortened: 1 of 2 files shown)


EmmanuelP:
Hi,

Sorry for being late reviewing this merge request. Be patient, I will dedicate some time to aravis this summer.

EmmanuelP:
Hi,

I have created a new pull request with some editions to your work: #919 

I will not change the default buffer size policy in the 0.8 stable series, but probably in the next development series.

As for the warning, I will not make it visible by default, as it is not a programming error. But I agree we should add an error placeholder to the set_recv_buffer function, and do something with it when an error is returned.

PR #909, opened 2024-06-24
Add missing private header.

feuerste:
The corresponding sources and headers were defined, but not the private headers.


EmmanuelP:
Thanks

Issue #915, opened 2024-08-14
Update features during runtime gstreamer


Summengardin:
**Describe the bug**
The "features" property are only set on the camera at creation. During runtime, the features are only updated in the local variable gst-aravis.

**To Reproduce**
- Create aravissrc-element
- During runtime, set property "features". For example Zoom

**Expected behavior**
To see the zoom level of the camera change. (Or other features for that matter)

**Camera description:**
Basler / TheImagingSource
a2A1920-51gcPRO / dfk 39gx265-z20
PoE (Ethernet)

**Platform description:**
 - Aravis version: 0.8.31
 - OS: Ubuntu 22.04
 - Hardware: Jetson

**Additional context**
The get-property method returns only the locally stored feature-string, so using that, everyhing seems fine.

Summengardin:
To set_property - case PROP_FEATURES, i added:
```
    			case PROP_FEATURES:
    			    GST_OBJECT_LOCK (gst_aravis);
			    g_free (gst_aravis->features);
                            gst_aravis->features = g_value_dup_string (value);
+                           if (gst_aravis->camera != NULL)
+				    arv_device_set_features_from_string (arv_camera_get_device (gst_aravis->camera), gst_aravis->features, NULL);
			    GST_OBJECT_UNLOCK (gst_aravis);
                            break;
```

PR #917, opened 2024-08-14
Issue/915/update-features-during-runtime

Summengardin:


Summengardin:
Fixes #915

EmmanuelP:
Thanks for the pull request.

I don't think it is a good idea. If you want to change the camera features, you can already get the plugin camera instance using the `camera` property. And by using it, you can handle gracefully any error that may happen when you set the feature.

I'm closing this pull request, unless you have a use case that can not be implemented using the `camera` property.

Issue #921, opened 2024-08-24
Aravis viewer endianness issue with mono 16-bit pixel format


berke:
Images are incorrectly displayed in Mono16 mode due to the endianness being wrong.

**To Reproduce**
- Run Aravis viewer
- Select Mono16 pixel format

**Expected behavior**
Image displays properly.

**Camera description:**
 - Manufacturer: Point Grey Research
 - Model: Blackfly BFLY-PGE-31S4M
 - Interface: GigE

**Platform description:**
 - Aravis version: 0.8.26-1 (Debian package), 0.10 (commit `2ebaa8661761...`)
 - OS: Debian 12
 - Hardware: x86_64

**Additional context**
- Changing `pgrPixelBigEndian` has no effect
- Changing the GStreamer format description from `GRAY16_LE` to `GRAY16_BE` in `arvmisc.c` solves the issue for me, but a proper solution may or may not require a kludge or an option in the UI to force endianness.
```
	{
		ARV_PIXEL_FORMAT_MONO_16,
		"video/x-raw, format=(string)GRAY16_BE",
		"video/x-raw",		"GRAY16_BE",
		"video/x-raw-gray, bpp=(int)16, depth=(int)16",
		"video/x-raw-gray",	16,	16,	0
	},
```

EmmanuelP:
Hi,

In the blackfly documentation, there is a `PixelBigEndian` feature. Isn't this feature set to False by default ?

EmmanuelP:
For the record, I have the same issue with a BFLY-PGE-14S2C-CS. It looks like an issue with the firmware, it seems they should send little endian pixels, unless asked for big endian.

<https://www.emva.org/wp-content/uploads/GenICam_SFNC_2_0_0.pdf>

```
12.2.66 GevSCPSBigEndian

Endianess of multi-byte pixel data for this stream.
This is an optional feature. A device that does not support this feature must support little-endian
and always leave that bit clear.
```

On my device, I have tried to set pgrPixelBigEndian to false and forced GevSCPSBigEndian , but without any effect on what is sent by the camera.

I'm closing this issue, as it is not aravis fault. Please reopen it if you don't think so.

berke:
> Hi,
> 
> In the blackfly documentation, there is a `PixelBigEndian` feature. Isn't this feature set to False by default ?

Thanks for checking on your side.  Yes when I turn the device on I get `pgrPixelBigEndian = false` and as I said it has no effect.

[...]

PR #922, opened 2024-08-26
Workaround for BlackFly U16 pixel endianness issues

berke:
Proposed solution for the endianness problem discussed here: https://github.com/AravisProject/aravis/issues/921

`ArvCamera` detects devices known to have pixel endianness issues using a model/vendor table, the result can be queried through `arv_camera_has_incorrect_pixel_endianness`.  This is used by the viewer to byte-swap pixels.

Shortcomings:
- It would be better to do the byte-swapping transparently in aravis as the packets are received and not in the viewer
- The user should be able to control the table of devices without having to recompile Aravis, or at least specify an option to activate the fix
- This code assumes the machine is little endian
- Code not optimized for the case where the rows have to be aligned AND the pixels byte-swapped
- Only handles MONO_U16

### DIFF
--- a/src/arvcamera.c
+++ b/src/arvcamera.c
@@ -122,6 +122,8 @@ typedef struct {
 
         gboolean has_region_offset;
 
+        gboolean has_incorrect_pixel_endianness;
+
 	GError *init_error;
 } ArvCameraPrivate;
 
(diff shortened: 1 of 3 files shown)


EmmanuelP:
Hi,

Thanks for your work.

I would not implement the byte swapping, gstreamer can do that for us, and possibly more efficiently. Also, your code will not work on big endian platform. I'm also not fond in exposing an API for a kludge.

Please move the camera list to arvmisc, and select the right gstreamer caps based on it there.

Emmanuel.

berke:
> Please move the camera list to arvmisc, and select the right gstreamer caps based on it there.

Done, see https://github.com/AravisProject/aravis/pull/923

PR #923, opened 2024-08-26
Workaround for BlackFly U16 pixel endianness issues

berke:
Continuation of https://github.com/AravisProject/aravis/pull/922

As requested, only arvmisc and arvviewer were modified, endianness is handled by the gstreamer plugin.

### DIFF
--- a/src/arvmisc.c
+++ b/src/arvmisc.c
@@ -804,30 +804,113 @@ ArvGstCapsInfos arv_gst_caps_infos[] = {
 	},
 };
 
+ArvGstCapsInfos arv_gst_caps_infos_incorrect_pixel_endianness[] = {
+	{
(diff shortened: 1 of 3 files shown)


EmmanuelP:
I have pushed a slightly edited version of your patch.

Thanks a lot.

berke:
Thank you, I've tested your edited version and it works as well.

BTW I'm also using aravis as a library, this fix only applies to the viewer and I still have to byte-swap the data in my code.  Not a big deal in this case, but IMHO it would have been better to handle this when the images are pushed to the ArvStream.

EmmanuelP:
I have deliberately avoided any data conversion in aravis library code. It is out of the scope of the library.

PR #925, opened 2024-08-27
gst: add stream property and notify on new property values

kohtala:
I need the statistics that are available on the stream object in a gstreamer pipeline. The stream was not accessible from the element.

To access the stream, I add a `stream` property. Since this property receives value at set_caps when the pipeline is prepared, I need to be notified of the new stream. For that I need to be able to connect to `notify::stream` signal.

I add the signal emission for both `camera` and `stream` properties as they both seem to change value by themselves.

### DIFF
--- a/gst/gstaravis.c
+++ b/gst/gstaravis.c
@@ -49,8 +49,7 @@ GST_DEBUG_CATEGORY_STATIC (aravis_debug);
 
 enum
 {
-  PROP_0,
-  PROP_CAMERA_NAME,
+  PROP_CAMERA_NAME = 1,
   PROP_CAMERA,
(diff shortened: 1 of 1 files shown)


EmmanuelP:
Hi,

I have created and applied to aravis-0-8 another pull request splitting your changes here: <https://github.com/AravisProject/aravis/pull/928>

The changes related to camera property notify signal is not included, as the logic for the detection of a camera change seems strange. You are comparing 2 pointers with one of them is freed.  There is nothing preventing the new allocated camera object is not at the same place as the one freed.

Please create a new pull request with only the notify:camera signal code, targeting aravis-0-8.

PR #931, opened 2024-09-02
Clang format

kohtala:
The source code has some lines indented with tabs and some lines indented with spaces expecting different tab widths.

This clang-format definition helps format code close to what I see the most code is formatted to.

Visual Studio Code has a [Clang-Format extension](https://marketplace.visualstudio.com/items?itemName=xaver.clang-format) that is able to make use of this definition, as I expect many other editors have their own plugins.

I did not test the workflow, but it may be close to something to check if PR is trying to push new badly formatted code. I find Codacy to also have some clang-tidy tool, but I am not familiar with that.


EmmanuelP:
Thanks.

Pushed with a tweak to preserve the consistency where a FALSE return value is often associated to an error.

PR #932, opened 2024-09-03
aravis-tools: Correct appstream metainfo description and file name

cmarmo:
Dear aravis maintainers,

this pull request fixes a [bug reported](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1078195) during debian packaging, applying the patch offered by @petterreinholdtsen.
It corrects appstream metainfo validation errors, adding missing `<p>` to description and making sure file name match ID.
Also it uses the new `metainfo.xml` ending replacing the obsolete `appdata.xml` ending.

Thanks for considering it.

### DIFF
--- a/viewer/data/meson.build
+++ b/viewer/data/meson.build
@@ -4,7 +4,7 @@ appdata_conf = configuration_data()
 appdata_conf.set ('ARAVIS_API_VERSION', aravis_api_version)
 
 configure_file (input: files('org.aravis.viewer.appdata.xml.in'),
-		output: 'org.aravis.viewer-@0@.appdata.xml'.format (aravis_api_version),
+		output: 'org.aravis.viewer.metainfo.xml'.format (aravis_api_version),
 		configuration: appdata_conf,
 		install_dir: aravis_app_data_dir)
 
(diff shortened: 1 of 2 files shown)


viewer/data/meson.build
```diff
@@ -4,7 +4,7 @@ appdata_conf = configuration_data()
 appdata_conf.set ('ARAVIS_API_VERSION', aravis_api_version)
 
 configure_file (input: files('org.aravis.viewer.appdata.xml.in'),
-		output: 'org.aravis.viewer-@0@.appdata.xml'.format (aravis_api_version),
+		output: 'org.aravis.viewer.metainfo.xml'.format (aravis_api_version),
```
EmmanuelP, line 7: Different aravis series are supposed to be installable in parallel. This change breaks this possibility.

EmmanuelP (changes requested):
Hi Chiara,

Thanks for the pull request.

The appdata filename change should keep the api version, and the component ID should be fixed to include it.

viewer/data/meson.build
cmarmo: Understood!
The last commit should have fixed this issue, right?

cmarmo: nope... I have to modify the id tag too... :) coming soon...

EmmanuelP: Not completely. In the bug report you mentioned, it is said the filename does not patch the component ID. I guess the `<id>` node should be modified, like what is done for the `<launchable>` node.

EmmanuelP (approved):

EmmanuelP:
Thanks Chiara.

PR #965, opened 2024-11-29
Fix unaligned access in tests/genicam.c

matwey:
Currently, this function leads to `SIGBUS` crash at armv7l architecture without `CONFIG_ALIGNMENT_TRAP` kernel option.


EmmanuelP:
Thanks a lot.

PR #979, opened 2025-02-05
Add gv: option to select discovery network interface

### LINKED ISSUE #505: Selecting network interface to be used by Aravis
Running on macOS, when I work with GigE cameras, I always make sure that I’m connected with ethernet, but often, I forget to disconnect my WIFI interface. Despite my LAN being higher priority than the WIFI, invariably, I get timeouts and the camera stops working. My only option, is to disconnect WIFI, and restart the application. It would be more robust/reliable if there was a way to steer the communications through the LAN interface. Flir Spinaker SDK seems to allow this by allowing to select a specific network interface to use to connect to the camera.

yuhamag:
fix #505 

This pull request adds an option to specify the network interface for discovering gv cameras.

The current implementation sends camera discovery packets to all network interfaces installed on the machine.
In this PR, by adding an option “gv-discovery-interface”, it is possible to send camera discovery packets only to a specific interface.

We have several methods of implementing setter/getter and network interface matching, as follows.

Method1: arvnetwork.c
  We need to implement the network interface matching for each OS.

Method2: arvgvinterface.c
  We can implement the minimum in the inherited class, but we cannot use setter/getter in other inherited classes.

Method3: arvinterface.c + arvgvinterface.c
  We can implement a setter/getter in the parent class and call the getter from the inherited classes.

We have implemented this PR using method3. What do you think?

@EmmanuelP

### DIFF
--- a/src/arvgvinterface.c
+++ b/src/arvgvinterface.c
@@ -490,7 +490,7 @@ static void
 arv_gv_interface_discover (ArvGvInterface *gv_interface)
 {
         int flags = arv_interface_get_flags (ARV_INTERFACE(gv_interface));
-	const char *discovery_interface = arv_interface_get_discovery_option (ARV_INTERFACE (gv_interface));
+	const char *discovery_interface = arv_interface_get_discovery_interface_name (ARV_INTERFACE (gv_interface));
 
 	_discover (gv_interface->priv->devices, NULL, flags & ARV_GV_INTERFACE_FLAGS_ALLOW_BROADCAST_DISCOVERY_ACK, discovery_interface);
 }
@@ -588,7 +588,7 @@ arv_gv_interface_camera_locate (ArvGvInterface *gv_interface, GInetAddress *devi
 		g_list_free_full (ifaces, (GDestroyNotify) arv_network_interface_free);
 	}
 
-	discovery_interface = arv_interface_get_discovery_option (ARV_INTERFACE (gv_interface));
+	discovery_interface = arv_interface_get_discovery_interface_name (ARV_INTERFACE (gv_interface));
 	socket_list = arv_gv_discover_socket_list_new (discovery_interface);
 
 	if (socket_list->n_sockets < 1) {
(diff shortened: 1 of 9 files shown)


EmmanuelP:
Hi,

Thanks for your work.

I don't think it is interesting to implement a generic ArvInterface API. What would this option mean to the other interfaces ? 

It would be better to rename `set/get_discovery_option` to `set/get_discovery_interface_name` to better reflect what they do.

Emmanuel.

yuhamag:
Hi,  

Thank you for your comment. I renamed the function name.

Currently, I don't have a specific meaning for other interfaces.
It would be better to move the setter/getter from ArvInterface to ArvGvInterface, right?

@EmmanuelP

EmmanuelP:
> It would be better to move the setter/getter from ArvInterface to ArvGvInterface, right?

Right.

src/arvgvinterface.c
```diff
@@ -353,8 +360,20 @@ struct _ArvGvInterfaceClass {
 
 G_DEFINE_TYPE_WITH_CODE (ArvGvInterface, arv_gv_interface, ARV_TYPE_INTERFACE, G_ADD_PRIVATE (ArvGvInterface))
 
+void
```
EmmanuelP, line 363: Make these two functions use arv_gv_interface_get_instance to find the interface instance. Use the mutex to protect private data access.

src/arvgvinterface.c
```diff
@@ -353,8 +360,20 @@ struct _ArvGvInterfaceClass {
...
+arv_gv_interface_set_discovery_interface_name (ArvInterface *interface, const char *discovery_interface)
+{
+	ARV_GV_INTERFACE (interface)->priv->discovery_interface = strdup (discovery_interface);
```
EmmanuelP, line 366: If discovery_interface was already set, a memory leak occurs. Free the content before setting the new one. Use g_strdup instead of strdup.

src/arvsystem.c
```diff
@@ -260,6 +260,29 @@ arv_set_interface_flags(const char *interface_id, int flags)
 	g_warning ("[Arv::enable_interface] Unknown interface '%s'", interface_id);
 }
 
+/**
```
EmmanuelP, line 263: Not needed.

src/arvtool.c
```diff
@@ -802,6 +808,9 @@ main (int argc, char **argv)
         if (arv_option_gv_allow_broadcast_discovery_ack)
                 arv_set_interface_flags ("GigEVision", ARV_GV_INTERFACE_FLAGS_ALLOW_BROADCAST_DISCOVERY_ACK);
 
+	if (arv_option_gv_discovery_interface)
+		arv_set_interface_discovery_option ("GigEVision", arv_option_gv_discovery_interface);
```
EmmanuelP, line 812: Use arv_gv_interface_set _discovery_interface_name() here.

yuhamag:
Thanks. I created the new commit regarding the comments.

@EmmanuelP

EmmanuelP:
I have committed a slightly edited version of your work.

Thanks !

PR #987, opened 2025-02-14
Add SPDX license identifier

### LINKED ISSUE #984: license mismatch (LGPL 2.1 vs 2.0)
as pointed out by @EmmanuelP in [this comment](https://github.com/microsoft/vcpkg/pull/42351#discussion_r1956041218) aravis is licensed as `LGPL-2.0-or-later`, however the `COPYING` file currently states that it's `LGPL-2.1-or-later`:
https://github.com/AravisProject/aravis/blob/ede6a5a46cb252741b7da59b2d897e561033d17c/COPYING#L2

there's also this file which states 2.1:
https://github.com/AravisProject/aravis/blob/ede6a5a46cb252741b7da59b2d897e561033d17c/viewer/arv-viewer.ui#L10

those were the only occurrences of 2.1 which i could find:
https://github.com/search?q=repo%3AAravisProject%2Faravis+%22version+2.1+of+the+license%22&type=code

may i suggest that you use the [SPDX license identifier](https://spdx.dev/learn/handling-license-info/) (e.g. in the README, though it's common to use it as the shorthand license header in all files instead of the big blob) as per [the SPDX license list](https://spdx.org/licenses/)? this makes it easier to immediately see which license it is (and it removes ambiguity around `LGPL-2.0-only` and `LGPL-2.0-or-later` & co.) + it helps automated scanners to identify the license

EmmanuelP:
Fix #984

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,14 +1,13 @@
-                  GNU LESSER GENERAL PUBLIC LICENSE
-                       Version 2.1, February 1999
+                  GNU LIBRARY GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
 
- Copyright (C) 1991, 1999 Free Software Foundation, Inc.
+ Copyright (C) 1991 Free Software Foundation, Inc.
  51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
-[This is the first released version of the Lesser GPL.  It also counts
- as the successor of the GNU Library Public License, version 2, hence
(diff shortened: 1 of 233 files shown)


COPYING
```diff
@@ -1,14 +1,13 @@
-                  GNU LESSER GENERAL PUBLIC LICENSE
-                       Version 2.1, February 1999
+                  GNU LIBRARY GENERAL PUBLIC LICENSE
```
rursprung, line 1: @EmmanuelP: same comment as here: https://github.com/AravisProject/aravis/pull/985/files#r1956751674
this now states "library" instead of "lesser" - i don't think that this was your intention?

PR #1002, opened 2025-03-28
gentl: always call IFUpdateDeviceList before IFOpenDevice

### LINKED ISSUE #1001: Cannot open Alkeria cameras
**Describe the bug**
Alkeria cameras cannot be opened by any tool offered by Aravis. Executing the viewer shows the camera correcly, but after selecting the camera the viewer will close with an error. These are the logs:

During enumeration:
> [15:52:40.222] 🅸 interface> TLOpen: /opt/alkeria/genicam/libalkusb3-genicam.cti
> [15:52:40.223] 🅸 interface> TLOpenInterface 'AlkUSBInterface'
> [15:52:40.241] 🅸 interface> Device: SOMESERIAL
> [15:52:40.242] 🅸 interface>   ID: SOMESERIAL
> [15:52:40.242] 🅸 interface>   VENDOR: Alkeria
> [15:52:40.242] 🅸 interface>   MODEL: CELERA C5S-MP
> [15:52:40.242] 🅸 interface>   S/N: SOMESERIAL
> [15:52:40.242] 🅸 interface> IFClose 'AlkUSBInterface'
> [15:52:40.434] 🅸 interface> TLClose: /opt/alkeria/genicam/libalkusb3-genicam.cti

After selecting the camera:
> [15:52:44.301] 🅸 interface> TLOpen: /opt/alkeria/genicam/libalkusb3-genicam.cti
> [15:52:44.301] 🅸 interface> TLOpenInterface 'AlkUSBInterface'
> [15:52:44.301] 🅸 interface> IFOpenDevice: 'SOMESERIAL'
> [15:52:44.302] 🆆 interface> IFOpenDevice: error -1007
> [15:52:44.302] 🆆 interface> DevClose error -1006
> [15:52:44.302] 🅸 interface> IFClose 'AlkUSBInterface'
> [15:52:44.302] 🅸 interface> TLClose: /opt/alkeria/genicam/libalkusb3-genicam.cti

Similar logs can be obtained by the other tools.

The issue is caused by a missing call to IFUpdateDeviceList before IFOpenDevice.

**To Reproduce**

1) connect the camera
2) execute any aravis tool
3) an error will be reported during a call to IFOpenDevice

**Expected behavior**
Camera should be opened without errors.

**Camera description:**
 - Manufacturer: Alkeria
 - Interface: Custom

**Platform description:**
 - Aravis version: both master and 0.9.0
 - OS: Ubuntu 20.04
 - Hardware: x86_64

**Additional context**
The GenTL 1.6 specification states that:

> In case the instantiation of a Device module is possible without having an internal device list, the IFOpenDevice may be called without calling IFUpdateDeviceList before.

The conditional statement implies that there are some producers which do not support calling IFOpenDevice without previously calling IFUpdateDeviceList, which is our case. So it should be sufficient to add this call to fix the issue, given that it will not affect other producer implementations.

psoftware:
Fixes #1001.

I am not sure this is the right way to fix it. In arvgentlsystem.c I see there is a *device* hashmap which is updated by the `_discover` function. That function invokes TLOpen -> IFOpen -> IFOpenDevice -> IFClose -> TLClose and keeps the hashmap updated with the current internal list of the gentl producer. Introducing this patch would allow to have a not up-to-date hashmap.

### DIFF
--- a/src/arvgentlsystem.c
+++ b/src/arvgentlsystem.c
@@ -363,6 +363,13 @@ arv_gentl_system_open_device_handle(ArvGenTLSystem *system, const char *interfac
 	/* Get interface handle */
 	interface_handle = arv_gentl_system_open_interface_handle(system, interface_id);
 
+	arv_info_interface("IFUpdateDeviceList");
+	error = gentl->IFUpdateDeviceList(interface_handle, NULL, 100);
+	if (error != GC_ERR_SUCCESS) {
+		arv_warning_interface("IFUpdateDeviceList: error %d", error);
+		return NULL;
+	}
+
 	arv_info_interface("IFOpenDevice: '%s'", device_id);
 	error = gentl->IFOpenDevice(interface_handle, device_id, DEVICE_ACCESS_CONTROL, &device_handle);
 	if (error != GC_ERR_SUCCESS) {


henriquesimoes:
@xiaoqiangwang, just pinging you since you first implemented the support for GenTL and might not be watching the repository. ;)

xiaoqiangwang:
I thought of that but  decided it would violate the "do-one-thing" principle. There is a public API, *arv_update_device_list* for that.

EmmanuelP:
I've just had a quick look. This patch is simple, but does not feel right. I think it would be better to make aravis not call IFClose and TLClose automatically on the last use. It would also probably improve the performance on device opening.

psoftware:
> I thought of that but decided it would violate the "do-one-thing" principle. There is a public API, _arv_update_device_list_ for that.

Surely it does not follow that rule, which also bothered me. Maybe this requirement should be moved above the interface.

> I've just had a quick look. This patch is simple, but does not feel right. I think it would be better to make aravis not call IFClose and TLClose automatically on the last use. It would also probably improve the performance on device opening.

It was the easiest way to avoid changing things at the caller site. Consider also that the internal implementation does not feel right too. The internal device hashmap cache is updated by the `_discover` function, which performs the opening of TLSystem, TLInterface, Device, etc. and closing all of them at the end, which makes all entries invalid, unless the caller performs a scan first before opening any of the scanned devices.

I will be able to work on this starting from the next week if you have any ideas (: Thank you for your reviewing!

PR #1017, opened 2025-05-28
arv_gc_node_set_integer_value. error to local_error

vladIsTrue:
It seemed strange to me to check the local_error variable for NULL when no one was working with it. I think something like this was meant


EmmanuelP:
Good catch. There is the same error in the function above.

PR #1024, opened 2025-07-11
gvsp: replace magic 0 with explicit constants

weypro:
Using the named constants clarifies intent, avoids magic numbers and improves type safety.

### DIFF
--- a/src/arvgvspprivate.h
+++ b/src/arvgvspprivate.h
@@ -107,6 +107,7 @@ typedef enum {
  */
 
 typedef enum {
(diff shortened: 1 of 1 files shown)


src/arvgvspprivate.h
```diff
@@ -107,6 +107,7 @@ typedef enum {
  */
 
 typedef enum {
+        ARV_GVSP_CONTENT_TYPE_NO_DATA =         0x00,
```
EmmanuelP, line 110: Please rename it to ARV_GVSP_CONTENT_TYPE_UNKNOWN for consistency

EmmanuelP (changes requested):

PR #1034, opened 2025-09-02
Fix non-zero offset ubs runtime error.

feuerste:
This fixes an `applying non-zero offset to null pointer` runtime error.
To reproduce:
```
env CC=clang CXX=clang++ meson setup -Db_sanitize=address,undefined -Db_lundef=false build
cd build
ninja
tests/fake
```
will give
```
../src/arvgcpropertynode.c:60:1: runtime error: applying non-zero offset 18446744073709551504 to null pointer
```
before the fix.


EmmanuelP:
Thanks

PR #1035, opened 2025-09-02
Fix left shift ubs runtime error.

feuerste:
To reproduce:
```
env CC=clang CXX=clang++ meson setup -Db_sanitize=address,undefined -Db_lundef=false build
cd build
ninja
tests/fakegv
```
will give
```
../src/arvgvdevice.c:2069:52: runtime error: left shift of 1 by 31 places cannot be represented in type 'gint' (aka 'int')
```
before the fix.


EmmanuelP:
Tanks

PR #1042, opened 2025-10-21
Add address display to viewer.

feuerste:
This way we can distinguish between GenTL and GigEVision.


feuerste:
Thanks a lot Emmanuel for all the information!

Unfortunately I don't have enough time at the moment for the proper implementation, so I'll just close this PR for now.

PR #1043, opened 2025-10-28
Add camera functions for exposure time and gain increment.

feuerste:
This is in particular useful for a Basler acA1920-25gc, where arbitrary exposure times can be set and read, but only multiples of the correct increment are actually applied by the camera driver.

Also the viewer has been adjusted to only allow multiples of the increment.

This was tested with an Allied Vision Alvium G1-1240c, FLIR Blackfly S BFS-PGE-23S3C, and 3 different Basler cameras.

### DIFF
--- a/viewer/arvviewer.c
+++ b/viewer/arvviewer.c
@@ -292,7 +292,7 @@ arv_viewer_value_to_log (double value, double min, double max)
 static double
 arv_viewer_value_from_log (double value, double min, double max, double inc)
 {
-	double result;
+	double power, mod, result;
 
 	if (min <= 0.0 || max <= 0)
 		return 0.0;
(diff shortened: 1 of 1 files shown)


src/arvcamera.c
```diff
@@ -1745,6 +1745,48 @@ arv_camera_get_exposure_time_bounds (ArvCamera *camera, double *min, double *max
...
 }
 
+/**
+ * arv_camera_get_exposure_time_increment:
+ * @camera: a #ArvCamera
+ * @error: a #GError placeholder, %NULL to ignore
+ *
+ * Returns: exposure time increment, in µs.
+ *
+ * Since: 0.10.0
```
feuerste: @EmmanuelP Is this the correct next version?

PR #1044, opened 2025-11-12
Add documentation link to bazel module.

feuerste:


feuerste:
@EmmanuelP I think it may be helpful to Bazel users to shortly mention its availability in the Bazel Central Registry. If you don't like this, just discard my PR :)

EmmanuelP:
Hi @feuerste ,

Thanks for your work on this, but I prefer not to mention bazel as an alternative build system. I will not use it, and don't want to handle possible bazel related issues. Bazel users are likely to find aravis by themselves in the central registry.

PR #1059, opened 2026-03-12
Add Gst 10/12/14/16-bit support

kohtala:
GStreamer 1.24 added support for more bit depths into bayer2rgb.

I only had Basler acA2440-20gc that implements BayerRG12. I tested on Testimage3 (for all pixel values) and Testimage6 (for correct colors) with command

```
gst-launch-1.0 -v aravissrc camera-name=192.168.0.20 ! video/x-bayer,format=rggb12le,framerate=1/1 ! \
  bayer2rgb ! video/x-raw,format=ARGB64_LE ! \
  videoconvert ! openjpegenc ! image/jp2 ! \
  filesink location=test-image-6.jp2
```

I also noticed a small potential leak in error path.

### DIFF
--- a/src/arvmisc.c
+++ b/src/arvmisc.c
@@ -720,9 +720,89 @@ ArvGstCapsInfos arv_gst_caps_infos[] = {
 		"video/x-raw-bayer, format=(string)bggr, bpp=(int)8, depth=(int)8",
 		"video/x-raw-bayer",	8,	8,	ARV_MAKE_FOURCC ('b','g','g','r')
 	},
+	{
+		ARV_PIXEL_FORMAT_BAYER_GR_10,
+		"video/x-bayer, format=(string)grbg10le",
+		"video/x-bayer",	"grbg10le",
+	},
+	{
+		ARV_PIXEL_FORMAT_BAYER_RG_10,
+		"video/x-bayer, format=(string)rggb10le",
+		"video/x-bayer",	"rggb10le",
+	},
+	{
+		ARV_PIXEL_FORMAT_BAYER_GB_10,
+		"video/x-bayer, format=(string)gbrg10le",
+		"video/x-bayer",	"gbrg10le",
+	},
+	{
+		ARV_PIXEL_FORMAT_BAYER_BG_10,
+		"video/x-bayer, format=(string)bggr10le",
+		"video/x-bayer",	"bggr10le",
+	},
+	{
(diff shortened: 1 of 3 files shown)


src/arvmisc.c
```diff
@@ -972,10 +1053,11 @@ arv_pixel_format_to_gst_0_10_caps_string (ArvPixelFormat pixel_format)
 		return NULL;
 	}
 
+	caps_string = arv_gst_caps_infos[i].gst_0_10_caps_string;
 	arv_debug_misc ("[PixelFormat::to_gst_0_10_caps_string] 0x%08x -> %s",
```
EmmanuelP, line 1057: arv_debug_misc should already handle NULL string correctly, displaying '(null)'. No need to check for NULL here.

src/arvmisc.c
```diff
@@ -986,6 +1068,8 @@ arv_pixel_format_from_gst_0_10_caps (const char *name, int bpp, int depth, guint
 	g_return_val_if_fail (name != NULL, 0);
 
 	for (i = 0; i < G_N_ELEMENTS (arv_gst_caps_infos); i++) {
+		if (arv_gst_caps_infos[i].name_0_10 == NULL)
```
EmmanuelP: Please use g_strcmp0

EmmanuelP (changes requested):
Thanks a lot. Just 2 minor things to fix.

src/arvmisc.c
kohtala: I placed the NULL case into the previous condition. Seems better there. I also did not find a promise of the '(null)' display in the glib documentation and it seems glib has a USE_SYSTEM_PRINTF fallback in some cases. So I'd feel safer not relying on it on any code path where it could break a valid case.

kohtala: I checked further. gcc is too smart and it optimizes printf("%s", NULL) into a puts(NULL) call, which made me think it does not handle it. printf() in glibc itself does print out the (null). But the Glib meson tests for using the system printf don't test for NULL printing (null). Anyway, I believe this is a feature Glib maintainers tell to rely on even when it's not documented. Anyway, I think how it's in the new commit is cleaner.

EmmanuelP:
Thanks for your investigation about g_strdup_printf. In the absence of a corresponding test in glib test suite, you are right it is better to take care to not send a NULL string parameter.

EmmanuelP:
Thanks.

PR #1069, opened 2026-04-27
v4l2 fixes

dkogan:
Hello. Thanks for merging the last one. Here are a few more patches. After these:

- I can use aravis with the latest mrcam frontend see my YUYV webcam images
- Same with MONO_8 images from v4l2loopback
- `arv-tool-0.10 genicam -n /dev/video0` doesn't spit out gigabytes of spaces

Note that https://github.com/AravisProject/aravis/commit/a458b346a279eab40e4665dfea3bb825293c0864 and https://github.com/AravisProject/aravis/commit/bb7faf2b45502a702ca324d14871d2a5785f4394 definitely address a bug, but I'm not sure the interpretation I guessed was correct. So maybe these do not fix the bug correctly, but it's correct-enough to work with mono_8 images.

Thanks!

### DIFF
--- a/src/arvdomparser.c
+++ b/src/arvdomparser.c
@@ -147,6 +147,21 @@ arv_dom_parser_characters (void *user_data, const xmlChar *ch, int len)
 		ArvDomNode *node;
 		char *text;
 
+                // Skip leading whitespace
+                while(len > 0 &&
+                      (*ch == '\n' ||
+                       *ch == '\r' ||
+                       *ch == '\t' ||
+                       *ch == ' '))
+                {
+                        ch++;
+                        len--;
+                }
+                // If nothing is left, we're done
+                if(len == 0)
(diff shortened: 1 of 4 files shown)


dkogan:
I just added one more patch: https://github.com/AravisProject/aravis/pull/1069/commits/1405e6ba443b9e06c6bc040e7954e54fa289a3ad

This ignores leading whitespace in DOM node text and ignores notes that are ALL whitespace. I was debugging with `ARV_DEBUG=all:3`, and without this patch I would see countless warnings such as: `dom> [ArvDomNode::append_child] Can't append '#text' to 'IntReg'`, at least when looking at v4l2 cameras

EmmanuelP:
Thanks Dima. I have merged everything except the last commit about the DOM debug output. I prefer to keep it as is for now.

PR #1070, opened 2026-04-29
Add IDS GV_53FxLE_M to the legacy list

aparenti:
We have got a GV_53FxLE_M camera in the facility, we figured out that one also needs the be in the legacy list otherwise we cannot get any data out of it.


EmmanuelP:
Thanks

PR #1071, opened 2026-04-29
v4l bytes_per_line_data: round the size up to the next byte

dkogan:
Since my interpretation of bytes_per_line_data appears to be correct (a458b346a279eab40e4665dfea3bb825293c0864 was merged), this patch has the logic you intended to add.


EmmanuelP:
Thanks

PR #1073, opened 2026-05-07
gst: get/set UserOutputValue as user-output-value

sorki:
This is useful for toggling a GPIO output configured before launching gst pipeline. We use this to toggle external LEDs on our camera.

Not sure if this is something worth upstreaming or a more general escape hatch would be better.

### DIFF
--- a/gst/gstaravis.c
+++ b/gst/gstaravis.c
@@ -59,6 +59,7 @@ enum
   PROP_GAIN_AUTO,
   PROP_EXPOSURE,
   PROP_EXPOSURE_AUTO,
+  PROP_USER_OUTPUT_VALUE,
   PROP_GAMMA,
   PROP_H_BINNING,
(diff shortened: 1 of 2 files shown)


EmmanuelP:
Did you try the `features` property ?

```
aravissrc features="UserOuputValue=1"
```

sorki:
Yea I've noticed that but wasn't sure if it actually sets it during runtime as well as it uses
```
gst_aravis->features = ..
```
can try!

EmmanuelP:
> Yea I've noticed that but wasn't sure if it actually sets it during runtime as well as it uses

Sorry, I've missed this requirement. No it currently does not set features while the pipeline is running. But it should be pretty trivial to change that.

EmmanuelP:
And it would be more generic than setting `UserOutputValue`.

PR #1076, opened 2026-06-03
Fix gstaravis triggered sparse streams 0.8

0xa1da34ef:
Add sparse stream handling for aravissrc

Backport of the sparse aravissrc stream handling change for the 0.8 branch.

This adds explicit timing and timeout policy controls so aravissrc can wait for sparse or triggered image streams without treating every empty buffer pop as a fatal error.

For triggered cameras, it is valid that no frame is available for some time. In that case the stream can still be healthy, so this change allows aravissrc to keep waiting when requested.

The existing fixed-rate behavior stays the default. Sparse timing skips fixed framerate caps and latency reporting, and GstBaseSrc unlock can interrupt waits cleanly.

Tested:
- Built from source and tested on Debian 12.
- Verified `aravissrc` through GStreamer with a Basler a2A1920-51gcBAS camera in triggered/sparse acquisition mode.

### DIFF
--- a/gst/gstaravis.c
+++ b/gst/gstaravis.c
@@ -72,6 +72,8 @@ enum
   PROP_USB_MODE,
   PROP_STREAM,
   PROP_TRIGGER,
+  PROP_TIMEOUT_POLICY,
+  PROP_TIMING_MODE,
   N_PROPERTIES
 };
 
@@ -85,6 +87,8 @@ enum
 };
 
 static guint gst_aravis_signals[LAST_SIGNAL] = { 0 };
+
+#define GST_ARAVIS_WAIT_TIMEOUT_CHUNK_US 200000
 
 #define GST_TYPE_ARV_AUTO (gst_arv_auto_get_type())
 static GType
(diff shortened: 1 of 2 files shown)


EmmanuelP:
Hi,

Did you adapt the patch for the aravis-0-8 branch, or is it just a cherry-pick ?

0xa1da34ef:
Hi,

This is just a straight cherry-pick from the corresponding change on main. No additional adaptations were required for the aravis-0-8 branch.

EmmanuelP:
Ok. I'll do the backport myself then.

0xa1da34ef:
Hi,

It is not a pure cherry-pick. The patch is based on the same change from main, but I adapted it for the aravis-0-8 branch because the aravissrc implementation differs between the branches. The functionality and intended behaviour are the same as in the main patch: keep the existing fixed-rate/default behaviour unchanged, add timeout-policy and timing-mode, avoid fixed framerate/latency handling for sparse timing, and allow GstBaseSrc unlock to interrupt waits cleanly. The main branch contains newer timing and latency handling code, while 0.8 still relies on older source timing logic (including the existing trigger-source based behaviour). As a result, the backport required adapting the implementation to the older plugin structure rather than applying the patch unchanged.

Sorry for the confusion in my previous answer.

PR #1077, opened 2026-06-08
add add-reference-timestamp-meta to aravissrc

danrs:
Add add-reference-timestamp-meta property, mirroring the property of the same name on rtspsrc. When enabled, each buffer gets a GstReferenceTimestampMeta with reference caps "timestamp/x-ptp", carrying the unmodified device timestamp. Buffers without a device timestamp (zero) get no meta.

The property defaults to FALSE, as devices with unsynchronized clocks will not provide absolute time

### DIFF
--- a/gst/gstaravis.c
+++ b/gst/gstaravis.c
@@ -73,6 +73,7 @@ enum
   PROP_USB_MODE,
   PROP_STREAM,
   PROP_TRIGGER,
+  PROP_ADD_REFERENCE_TIMESTAMP_META,
   N_PROPERTIES
 };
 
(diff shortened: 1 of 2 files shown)


EmmanuelP:
Hi,

Thanks for the pull request. Could you explain the use case ?

My current comprehension is when you know a camera is ptp synchronized, you enable this feature that will add an additional caps to every gstreamer buffer using the aravis buffer timestamp field. Would it not possible to detect a camera is ptp configured and add the additional caps automatically ?

danrs:
I'm after high-accuracy frame-capture timestamps on my video frames. My cameras sync to PTP, but the PTP stamp is not propagated into the pipeline.

Your reading is correct, enabling this setting adds the ptp caps to every buffer. I have downstream consumers that make use of the timestamp.

I've kept the toggle as a manual (default false) setting to match rtspsrc. Furthermore, the user needs to configure PTP manually on the camera anyway, so auto-enable in the pipeline is of negligible benefit.

If it were to be done automatically, we'd be looking at:
- A blocking register read on the streaming thread (or adding a thread/timer)
- different behaviour to the existing reference timestamp property on rtspsrc
- a lookup table for PtpStatus property name/value on cameras from different manufacturers (eg. legacy cameras use `GevIEEE1588Status` with `SLAVE`, `MASTER`)

danrs:
@EmmanuelP any update or comment on this?

PR #1085, opened 2026-07-05
Tag device id with transport to avoid GEV/U3V id collision

circuitvalley:
A camera that reports the same vendor/model/serial over both GigE Vision and USB3 Vision produced an identical device id on each interface. Because arv_open_device() returns the first matching interface and USB3Vision is enumerated before GigEVision, selecting the GigE device silently opened the USB one.

Append a transport tag ("-eth" / "-usb") to the id built in each interface so the two are distinct and each resolves to the intended device.

### DIFF
--- a/src/arvgvinterface.c
+++ b/src/arvgvinterface.c
@@ -293,7 +293,12 @@ arv_gv_interface_device_infos_new (GInetAddress *interface_address,
                 infos->serial = g_strdup (infos->mac);
         }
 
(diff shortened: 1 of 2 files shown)


EmmanuelP:
Hi. Thanks for the patch.

Unfortunately, it is a breaking change. Also, it is already possible to select the right device, by using one of the other names it is register under.

For example, you can use the MAC address or the ip address of a GigEVision camera, or the USB guid for a USB3Vision device.

PR #1086, opened 2026-07-10
Add AnalogControl feature to Root category

TRYOKETHEPEN:
There are pFeatures like _GainRaw_ in XML, but _Root_ does not reference _AnalogControl_


EmmanuelP:
Thanks

PR #1087, opened 2026-07-22
gst: fix aravissrc teardown deadlock under incomplete frames

MrBlenny:
### Problem
Setting a pipeline containing `aravissrc` to `NULL` can hang indefinitely. It shows up when complete frames stop arriving - high frame rate, GigE packet loss, bandwidth contention from several cameras, or an unplugged camera and gets worse with more cameras / higher frame rate.

### Root cause
`gst_aravis_create()` waits for a buffer while holding `GST_OBJECT_LOCK`, and the element registers no `GstBaseSrc` `unlock()` vmethod, so the base class has no way to interrupt the wait during a state change. When the stream only yields incomplete (non-`SUCCESS`) buffers, the `do { … } while (status != SUCCESS)` loop spins on them forever and never releases the lock, so the `PAUSED→READY` transition deadlocks waiting for the streaming task to pause / for the lock.

### Fix
- Add `unlock()` / `unlock_stop()` that set/clear an atomic `flushing` flag.
- `create()` now takes a ref on the stream and **drops `GST_OBJECT_LOCK` across the buffer wait**, polls in short slices (`GST_ARAVIS_BUFFER_POLL_SLICE_US`) checking the flag and returning `GST_FLOW_FLUSHING` promptly on unlock.
- The wait is bounded to `buffer_timeout_us` total; on continuous failure buffers it now returns `GST_FLOW_ERROR` instead of spinning.
- `start()` clears the flag so a reused element streams again.

### Behaviour changes (intentional)
- Sustained incomplete frames now surface as `GST_FLOW_ERROR` after `buffer_timeout_us` rather than an unbounded spin.
- A missing stream at `create()` time returns `GST_FLOW_FLUSHING` instead of hitting an `ARV_IS_STREAM` `g_return_val_if_fail` CRITICAL + `GST_FLOW_ERROR`.

### Test
New `gstreamer` test suite (`tests/arvgstteardowntest.c`): an in-process fake GV camera with a high `gvsp-lost-ratio` feeds `aravissrc ! fakesink`; the test asserts `set_state(NULL)` completes within a bound. It aborts (deadlock) without the fix and passes with it. Built only when the plugin is enabled; excluded from the valgrind setup like the `python` suite. `subdir('gst')` now runs before `subdir('tests')` so the suite can link the plugin.

Note: I built and ran this against 0.9.3-dev in a container - meson test --suite gstreamer fails (SIGABRT) without the fix and passes with it; a normal 25 fps stream still delivers frames with clean EOS.

### DIFF
--- a/gst/gstaravis.c
+++ b/gst/gstaravis.c
@@ -47,6 +47,8 @@
 
 #define GST_ARAVIS_DEFAULT_N_BUFFERS		50
 #define GST_ARAVIS_BUFFER_TIMEOUT_DEFAULT	2000000
+/* Poll the buffer wait in slices this size so unlock() is noticed promptly. */
+#define GST_ARAVIS_BUFFER_POLL_SLICE_US		100000
 
 GST_DEBUG_CATEGORY_STATIC (aravis_debug);
 #define GST_CAT_DEFAULT aravis_debug
@@ -488,6 +490,10 @@ gst_aravis_start (GstBaseSrc *src)
         else
                 GST_LOG_OBJECT (gst_aravis, "Open first available camera");
 
+	/* Clear a stale flush from a previous cycle (unlock_stop is not guaranteed
+	 * on the way to NULL) so a reused element streams again. */
+	g_atomic_int_set (&gst_aravis->flushing, 0);
+
 	GST_OBJECT_LOCK (gst_aravis);
 	if (gst_aravis->camera == NULL)
 		result = gst_aravis_init_camera (gst_aravis, &notify, &error);
@@ -560,6 +566,28 @@ gst_aravis_get_times (GstBaseSrc * basesrc, GstBuffer * buffer,
 	}
 }
 
+/* Without unlock(), a create() blocked waiting for a buffer (no complete frames
+ * arriving) can't be interrupted, deadlocking the state change to NULL. */
+static gboolean
+gst_aravis_unlock (GstBaseSrc *src)
+{
+	GstAravis *gst_aravis = GST_ARAVIS (src);
+
+	g_atomic_int_set (&gst_aravis->flushing, 1);
+
+	return TRUE;
+}
+
+static gboolean
+gst_aravis_unlock_stop (GstBaseSrc *src)
+{
+	GstAravis *gst_aravis = GST_ARAVIS (src);
+
+	g_atomic_int_set (&gst_aravis->flushing, 0);
+
+	return TRUE;
+}
+
 static GstFlowReturn
 gst_aravis_create (GstPushSrc * push_src, GstBuffer ** buffer)
 {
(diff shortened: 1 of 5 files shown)


EmmanuelP:
Hi,

Thanks for the patch. Not an issue, but did you use AI to cook this one up ?

The regression test is welcome. But it fails for me:

```
test:         gstreamer - aravis:gst-teardown                                                                           
start time:   15:31:21                                                                                                  
duration:     19.16s                                                                                                    
result:       killed by signal 6 SIGABRT                                                                                
command:      MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 UBSAN_OPTIONS=halt_on_error=1:           ->abort_on_error=1:print_summary=1:print_stacktrace=1 LD_LIBRARY_PATH=/home/pacaud/Sources/aravis/build/src:/opt/MVS/lib/64:/opt/MVS/   ->lib/32:/opt/MVS/lib/64:/opt/MVS/lib/32 MESON_TEST_ITERATION=1 GST_PLUGIN_PATH=/home/pacaud/Sources/aravis/build/gst                   ->ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 MALLOC_PERTURB_=51 /home/pacaud/Sources/aravis/build/tests/arv-gst-     ->teardown-test
----------------------------------- stdout -----------------------------------                                          
TAP version 14                                                                                                          
# random seed: R02Sa7ec2811353a872f4888b1676c32288e                                                                     
# GLib-GIO-DEBUG: _g_io_module_get_default: Found default implementation gvfs (GDaemonVfs) for ‘gio-vfs’                
1..1                                                                                                                    
# Start of gstreamer tests                                                                                              
not ok /gstreamer/teardown-under-packet-loss - ERROR:../tests/arvgstteardowntest.c:77:teardown_under_packet_loss_test: 'torn_down'      ->should be TRUE
Bail out!                                                                                                               
----------------------------------- stderr -----------------------------------                                          
**                                                                                                                      
ERROR:../tests/arvgstteardowntest.c:77:teardown_under_packet_loss_test: 'torn_down' should be TRUE 
```

While running the test by hand works:

```
❯ GST_PLUGIN_PATH=./gst/ ./tests/arv-gst-teardown-test
TAP version 14
# random seed: R02S54e1b7b61d88edfc6b13953f4f5322b7
# GLib-GIO-DEBUG: _g_io_module_get_default: Found default implementation gvfs (GDaemonVfs) for ?gio-vfs?
1..1
# Start of gstreamer tests
ok 1 /gstreamer/teardown-under-packet-loss
# slow test /gstreamer/teardown-under-packet-loss executed in 1.05 secs
# End of gstreamer tests

```

 Any idea why ?

EmmanuelP:
CI pipelines all fail

